PR #24, opened 2026-05-20
feat: bind directory when creating groups

DeepColds:
## Summary
- add `--working-dir` / `--cwd` / `--dir` to `botmux create-group` so new groups can be bound to an oncall directory during creation
- propagate the same bind option through the dashboard Create new group flow and show per-bot bind results
- validate bind directories before chat creation and return per-bot binding status without retrying chat creation on partial bind failures

## Tests
- `pnpm vitest run test/group-creator.test.ts test/dashboard-ipc.test.ts`
- `pnpm build`


deepcoldy (approved):
Review 收尾已完成：目录校验统一复用 validateWorkingDir，修复 CLI 裸 ~ 展开 bug，构建与相关测试（124）通过。LGTM。

PR #40, opened 2026-05-22
Workflow runtime v0.1.0 → v0.1.6

sensuossss:
## Workflow runtime v0.1.0 → v0.1.6

Merges the entire workflow runtime line that's been baked under canary tags (v2.31.0-canary.* → v2.33.0-canary.5) plus a fresh v0.1.6 attempt-replay + manual-resume slice into `master`. Real-machine dogfood + dedicated test files green. Surface is **experimental / canary-grade** — see "Stability promise" below.

**Scope**: 198 commits since master · 226 files · +47.9k / -915 lines.

---

### Event runtime (v0.1.2 — frozen)

NDJSON `EventLog` per run, deterministic replay → in-memory snapshot, idempotency keys on every mutating event, schedule-store with deterministic ids. `cancelRequested` / `cancelDelivered` / `cancelOnDaemonOk` plumb the cancel propagation chain end-to-end. Schema versioned (`schemaVersion` on every event); recovery & reconcile worked through codex rounds 1–5.

### Recovery (R0 / R1)

- `R0` — daemon startup `cold-attach`: any active run on disk gets its event-watch + reconciler wired back up
- `R1` — CLI cold-start: `botmux workflow resume <runId>`, with `inputHash` guard rejecting tampered sidecars before any provider call
- Cold-attach E2E covered (`workflow-cold-attach.test.ts`); cancel × reconcile co-operation locked in

### Host executors (side-effect nodes)

`feishu-send`, `feishu-reply`, `botmux-schedule` registered + dogfooded with their own demo workflows. Ambiguous inputs rejected pre-flight. Output blobs spilled to disk when > inline cap.

### CLI surface

```
botmux workflow run | resume | cancel | ls | tail | validate
```

Every operation is event-log first; CLI is a thin entrypoint over the same orchestrator the IM and dashboard share. `ls` / `tail -f` plus the `--plain` table outputs are scripting-friendly.

### IM surface (Feishu)

- `/workflow run <id> --param k=v` from any chat / thread; cancel-scope is the source chat
- Approval cards: enforced `approvers` whitelist; approve / reject / cancel buttons (horizontal column_set layout, frozen on click)
- Long human-gate prompts auto-spill to blob, card shows `promptPreview`
- **Run-level progress card** (v0.1.5 slice 1): card lives in the originating thread, patched on every event (nodes done · failed · cancelled, running rows, waiting rows, failure summary). Cancel Run button surfaces straight on the card.
- **Per-attempt dashboard deeplink** (v0.1.5 slice 3): `[查看当前终端]` button on the progress card opens the live xterm for the running attempt in the dashboard

### Dashboard

- Run list (5s poll) + Run Detail (2s poll, stops at terminal)
- Summary, dangling-work red panel, node/activity rows, event timeline, **parallel-execution timeline** (attempt-level)
- **Cancel / approve / reject directly in UI** with reviewer comments
- **Workflow Catalog** — list / detail / dependency graph / trigger run with params input
- **Embedded per-attempt web terminal** with iframe caching across polls (no flicker)
- Decision auth lifted into a pure function (`decideDashboardAuth`) with HTTP-level smoke locking the public-read / write-protected matrix
- i18n zh/en + topbar locale + theme (system / light / dark) switch
- **Full dark mode adaptation** including workflow/catalog surfaces — all colours go through semantic tokens (`--accent`/`--success`/`--warning`/`--danger` + `*-soft` pairs), parallel timeline foreground inverted in dark to maintain contrast

### Workflow create skill (`botmux-workflow-create`)

Translates natural-language stepwise tasks into validated workflow JSON at `$HOME/.botmux/workflows/<id>.workflow.json`. Hard rules: `subagent.bot` MUST be `larkAppId` (not displayName); `$ref` for whole-value substitution only; `humanGate.stage: before`; high-risk side-effect nodes default to gated.

### Parallel dispatch (v0.1.3)

- `runLoop` batches dispatchable activities; `Promise.allSettled` with `maxConcurrency` cap (default 4)
- Per-bot tick-local serialization to avoid pounding a single CLI's chat-side state
- `pre-attempt dispatch throw` no longer silently retried; M2 dual-terminal invariant enforced

### Cancel responsiveness (v0.1.4-a)

- Orchestrator short-circuit on `cancelledRunIntent` — no more late dispatch racing the cancel
- `AbortController` / `AbortSignal` protocol threaded `runLoop → dispatchWork → spawn`; reason carries `cancelOriginEventId`
- Daemon-side `entry.aborters` map; `entry.cancelling` synchronous dedup against concurrent cancel races
- Worker spawn: `close` IPC + `SIGINT` + **5 s grace** + `SIGKILL` escalation; `cleanup({ skipSigterm })` keeps the cancel grace from racing the normal 250 ms SIGTERM
- **Cancel-wins** semantics: `quiesce` / `hardDeadline` disarmed on abort; `finish` / `fail` short-circuit if `cancelRequested` — `activityCanceled` fires only after worker `'exit'`
- `WorkerSpawnResult` adds `kind: 'cancelled'` with required `cancelOriginEventId`
- 200 ms polling fallback observer for external EventLog writes that bypass the daemon

#### Dogfood (real Claude/Codex CLI subprocesses)

| scenario | `cancelRequested → runCanceled` | worker SIGINT → exit |
|---|---|---|
| early cancel (single worker mid-sleep) | 25 ms | 11 ms (cooperative) |
| parallel cancel (two workers concurrent) | 19 ms | 10 ms (both) |

Both branches' `activityCanceled` link the same `cancelOriginEventId`. No SIGKILL needed in either run.

### Dogfood bug fixes (v0.1.5 hardening)

Caught during real-machine v0.1.5 canary runs and fixed inline:

- Dashboard workflow terminal iframe flicker — per-card surgical DOM update (iframe parent stays attached so browsing context isn't discarded)
- Progress card denominator: read `ctx.def.nodes` count for the true total instead of `snap.nodes.size` (grows over time)
- Concurrent PATCH race producing red→blue header flicker — per-runId `updateChain` promise serializer
- `running + failed` co-existence on terminal cards — clear running/waiting rows once run.status is terminal
- humanGate approval card: failed/cancelled fields split, approve/reject in `column_set` row, click freezes the card (returns no-form body for in-place patch), non-approver gets toast

### Compat fix — Claude CLI 2.1.146 UUID session-id

`syntheticIds.sessionId` is now a deterministic UUID-v4-shaped token (SHA-256 of the raw `wf-<runId>-<activityId>-<attemptId>` rewritten to satisfy the v4 shape). Claude Code 2.1.146 started rejecting `--session-id` strings that aren't valid UUIDs ("Invalid session ID. Must be a valid UUID."); before this fix, every workflow worker crashed on spawn.

### Attempt replay + manual resume (v0.1.6)

Once a node finishes, the worker SIGTERMs 250 ms after `final_output` (cancel grace is still untouched). Before v0.1.6 the live xterm iframe just disconnected; the operator could only inspect a 100 KB plain-text tail in the run-detail page. v0.1.6 brings two new operator surfaces over the same `terminal.log` sidecar:

- **xterm.js replay** — `GET /api/workflows/runs/:rid/attempts/:aid/:atid/terminal-log/raw` streams the raw bytes (tail 10 MB default; `?tailBytes=N` up to 64 MB; `?download=1` attachment up to 512 MB), with `X-Botmux-Log-Bytes` / `X-Botmux-Served-Bytes` / `X-Botmux-Truncated` headers. A self-contained `assets/terminal-replay.html` (xterm.js CDN, no bundled deps) chunk-writes the response (64 KB / `requestAnimationFrame`) so a multi-MB log renders without freezing the tab. ANSI colours / cursor moves all replay correctly.
- **Manual resume (debug session)** — operator clicks "Resume session" on a terminal attempt → daemon spawns a fresh CLI worker with `--resume <cliSessionId>`, the dashboard iframe swaps to the new worker's write-enabled PTY URL (`http://host:port?token=<writeToken>`), and the operator can keep chatting with that CLI instance. Resume workers are **dashboard-only debugger sidecars**: they never write to the workflow EventLog, never advance run state, never appear as a new attempt. Lifecycle is attempt-scoped singleton (re-click returns the same `webPort`, no double-spawn), idle 30 min auto-kill, explicit "End debug session" button. Resume terminal output lands in `attempts/<activity>/<attempt>/resumes/<resumeId>/terminal.log` so the immutable original log stays clean for replay viewers.

Adapter capability gate: Claude Code / Codex / CoCo / Aiden / Antigravity / Cursor honor `--resume <sessionId>` with the captured `cliSessionId` (B0 added `cliSessionId` to `WorkerSessionInfo` + `AttemptTerminalSidecar` and threads it through worker IPC so daemon picks up adapter-rotated session ids). Opencode / Gemini are explicitly disabled — Opencode advertises no resume support, Gemini's `--resume` only takes "latest" and can't deterministically target an attempt session.

Security: replay endpoint is public-read (consistent with existing input/output snapshot exposure); start/end resume are write-protected via the existing `decideDashboardAuth` cookie boundary. Write-token PTY URLs only land inside authenticated dashboard sessions.

Slice map (all live under `feat/workflow-runtime`):
- A1 — raw terminal-log endpoint (path traversal guard + tail/download)
- A2 — replay HTML + xterm.js chunk write + dashboard slot switching
- B0 — `cliSessionId` persistence (final adapter-native id, not spawn-time only)
- B1 — daemon `AttemptResumeManager` (Map<run/activity/attempt> single-lock, sidecar files, write-token web terminal, SIGINT → 5 s grace → SIGKILL teardown)
- B2 — dashboard "Resume session" button + write-token iframe (with CLI capability tooltip + missing-cliSessionId guard)

---

## Testing

- **Workflow targeted**: 41+ test files, all green
  - `workflow-daemon-spawn` 13 · `workflow-spawn-bot` 16 · `workflow-spawn-cancel` 5 · `workflow-cancel-responsiveness` 6 · `workflow-cancel-finalize-e2e` 1 · `workflow-approval-card` 13 · `workflow-progress-card` 24 · `workflow-canary` 1 · `workflow-runtime` (orchestrator/replay/reconcile/dispatch) · dashboard API smoke · IM cancel · cold-attach · params binding · output binding · host executors · etc.
- **Workflow + Dashboard + Card + Session subsets**: 911 tests · all green on `f3708d0` (post-merge baseline)
- **v0.1.6 add-on**: `workflow-attempt-resume` (16) + new `dashboard-workflow-api` raw-log subtests (6); workflow + dashboard subset 133/133 on `ce287da`. `tsc --noEmit` + `pnpm build` clean
- **Codex independent review**: rounds 1–6 during v0.1.4-a + v0.1.5 slices + dark-mode QA + v0.1.6 design rounds (idle-watchdog decision deferred, replay + resume split LGTM). Final LGTM with the hygiene blockers cleaned in canary tags
- **Real-machine dogfood**: `cancel-dogfood.workflow.json` driven through early-cancel + parallel-cancel scenarios with real Claude Code subprocesses; numbers above. v0.1.6 replay + resume currently green in unit tests; bench-style dogfood is the natural next step on top of this merge

## Stability promise

Workflow runtime is **experimental / canary surface** for now. Event schema is intended to stay backward-compatible (replay-safe additions only), and host-executor input contracts are stable inside this PR — but cross-major persistent compatibility is not promised yet. Treat the public CLI / IM / dashboard verbs as stable; treat internal types (`WorkerSpawnFn`, `WorkflowRuntimeContext`, etc.) as fair game to evolve.

## Known follow-ups (NOT blockers for this PR)

- **Workflow fanout crash window**: watcher delivery is at-least-once while the daemon stays alive, but there is still a tiny crash-during-send window (event observed, card/send side effect in flight, daemon dies before any durable delivery ack). Restart will reattach the run, but it cannot prove whether that side effect landed, so it will not auto-recover the fanout message. Follow-up shape: durable fanout outbox / delivery ack.
- **v0.1.4-b**: configurable `cancelPolicy` in node schema (`kill` / `signal` / `inject-prompt` opt-in; compensation modeled as separate events)
- **v0.1.4 L2**: replace the 200 ms cancel polling observer with an `EventLog` emitter — purely a cost optimization
- **Claude Bash idle vs final_output**: Claude Code's Bash tool can voluntarily background long commands and emit an interim "waiting…" assistant message; the workflow runtime today treats that as `final_output` and rejects the attempt. Current workaround is on the workflow design side (force foreground via continuous-stdout shell loops). Runtime fix candidate: require a `<WORKFLOW_OUTPUT>` marker to be present before accepting `final_output`. File a follow-up issue post-merge.
- **Per-node `timeoutMs` UX**: schema already supports it (`node.timeoutMs` / `defaults.timeoutMs`), but daemon-spawn still falls back to a 5-min wall-clock cap when neither is set. Discussion decided to drop that default (run unlimited unless author opts in) + have `botmux-workflow-create` propose recommended values per node type. Not in this PR — wire on a follow-up.
- **Idle watchdog (vs wall-clock)**: separate PTY-activity heartbeat that fails workers stuck silent for N minutes, decoupled from the absolute wall-clock cap. Designed with codex (heartbeat-throttled IPC, daemon-side timer, fail-not-cancel teardown) but punted per operator call.

## Merge strategy

Squash merge — 198 commits roll up into one master commit. Branch is `feat/workflow-runtime`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


deepcoldy:
## Review 结论：通过，合并 🎉

感谢 @sensuossss 这一轮扎实的打磨 —— 上一轮 review 提的问题基本全部落地，而且解法都很到位。

### 已修复并验证
- **R-M1**：冷启动 `attachColdWorkflowRuns` 现包了 try/catch，cold-scan 失败只 warn、不再阻塞 scheduler 启动。
- **C1/C2**：fanout `drain()` 出错改为 early-return 不前进游标（同进程 at-least-once 重投），并新增 5s poll 定时器兜底 fs.watch。验证过：不会死锁、不会短路后续 drain、正常成功路径不会重复发卡。
- **B3**：humanGate 从软约定升级为 schema 硬规则 —— side-effect executor（`feishu-send`/`feishu-reply`/`botmux-schedule`）无 `humanGate` 且无显式 `unsafeAllowUngated` 会在 parse 阶段直接拒绝；仓库 demo 也全部改正。
- 附带：worker 关闭 SIGTERM→SIGKILL 升级兜底；bridge watcher 的 `cliSessionId`/`sessionId` 不一致修复。

质量：`tsc --noEmit` 干净；合并最新 master 后全量 2396 测试通过（仅 e2e-browser 因缺少真实飞书环境变量跳过）。

### 已知并接受（beta 姿态，非阻塞）
1. **pty.log 公开读**：`GET /api/workflows/.../terminal-log/raw?stream=pty` 命中 `/api/workflows/*` 公开读前缀，pty.log 是原始全终端录像、敏感度高于此前的 `terminal.log`。beta 阶段接受该 public-read 边界扩大，暂不加 token gate。
2. **fanout 跨重启 at-least-once 缺口**：同进程重投已可靠，但冷启动 `lastSeq=currentSeq()` 从 EOF 开始，daemon 若在「waitCreated 已落盘、审批卡未发成功」窄窗口崩溃，重启后不会补发该审批卡。接受为 **known-limitation** —— 建议在 `fanout.ts` 补一行注释说明，且不对外宣称「fanout 已完整闭环」；彻底解决需持久 delivery ack 或冷启扫描 open waits 补发。

### 小维护点
`SIDE_EFFECT_EXECUTORS` 需手动与 dispatch 表保持同步，将来新增 side-effect executor 别忘了加进来，否则会默认不 gate。

无阻塞项，合并。再次感谢 🙏

deepcoldy (approved):
Review 通过，详见上方评论。无阻塞项，合并。感谢 @sensuossss 🙏

PR #58, opened 2026-05-27
fix(hermes): bridge fallback replies from state db

Hellovolcanoone:
fix(hermes): bridge fallback replies from state db


deepcoldy (approved):
Reviewed in isolated worktree + independent second review by Codex.

- Build green; full unit suite 2687 passed (the 7 failures in bridge-final-output-retry.test.ts are pre-existing on master — stale resolveBrandLabel mock — unrelated to this PR).
- Original blocker (cross-session leak: global ~/.hermes/state.db read with no session filter) is resolved in a896363: assistant_final now only closes a turn whose sourceSessionId (Hermes native session_id) matches, and only finish_reason='stop' rows are treated as final. New guard test added.
- Note for a real-Hermes smoke test: the protection is inert if messages.session_id is not populated per row; worth confirming once on live Hermes.

PR #101, opened 2026-06-03
Clarify botmux send mention switches in built-in skill

LucasIcarus:
Bot send msg like this sometimes
image
And the analysis result listed below
image

strengthen the internal botmux send skill a little bit. 

## Summary
- Clarify in the built-in `botmux-send` skill that `--mention-back` and `--no-mention` are value-less switches.
- Point agents to `--mention <open_id:name>` for explicit recipients, and document the existing content-source precedence for multiline bodies.
- Add regression coverage for the built-in skill guidance.

## Test plan
- `pnpm vitest run test/builtin-skills.test.ts test/send-policy.test.ts`
- `pnpm build`


deepcoldy (approved):

deepcoldy:
看着没问题，就不拉机器人review了，合入了人工测试下就发版

PR #150, opened 2026-06-08
feat(card): support manual working directory selection

fyydnz53:
## Summary

- add a manual working-directory input to the repository selection card
- allow the pending session owner to submit an existing local directory directly from the card
- start or switch the session in that selected directory
- add card-builder and card-action coverage for the manual submission flow

## Why

Project scanning may not list every useful working directory. Users can already type `/repo <path>` manually; this adds the same direct-path workflow to the interactive card.

## Validation

- `pnpm test -- test/card-builder.test.ts test/repo-manual-card-action.test.ts`
- `pnpm build`


deepcoldy:
@fyydnz53 这个PR是已经完成了吗，我开始review？

PR #271, opened 2026-06-22
fix: isolate tmux paste buffers

nil-err:
## Summary
- use unique named tmux buffers for paste operations in both tmux backends
- delete named paste buffers after use to avoid cross-session default-buffer races
- add unit coverage for named-buffer behavior and fresh buffers per paste

## Verification
- corepack pnpm install --frozen-lockfile
- corepack pnpm vitest run --project unit test/tmux-pipe-backend.test.ts test/tmux-backend-input.test.ts test/tmux-copy-mode.test.ts
- corepack pnpm exec tsc --noEmit
- git diff --check

### DIFF
--- a/src/adapters/backend/tmux-backend.ts
+++ b/src/adapters/backend/tmux-backend.ts
@@ -2,6 +2,7 @@ import * as pty from 'node-pty';
 import { execSync, execFileSync } from 'node:child_process';
 import { accessSync, constants as fsConstants } from 'node:fs';
 import { basename } from 'node:path';
+import { randomBytes } from 'node:crypto';
 import type { SessionBackend, SpawnOpts, SessionProbe } from './types.js';
 import { probeTmuxFunctional, tmuxEnv } from '../../setup/ensure-tmux.js';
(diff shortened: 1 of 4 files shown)


deepcoldy:
Thanks a lot for this fix, @nil-err! 🙏

This was a genuine cross-session bug: botmux's tmux paste path loaded into the server-global **default** buffer, so when multiple worker processes pasted concurrently, worker B's `load-buffer` could slip between worker A's `load-buffer` and `paste-buffer` — and A would paste B's text into the wrong pane (content cross-talk). Switching to a per-paste uniquely-named buffer (`botmux-<hex>`) closes that window completely, and the `finally{}` guard keeps named buffers from accumulating when a paste fails. Nice, surgical fix.

Reviewed + independently verified (tsc clean, targeted tests green, and confirmed these are the only runtime tmux load/paste paths so there's no residual race). I layered on a few extra unit tests before merging — a symmetric fresh-buffer test for `TmuxBackend` and a buffer-cleanup-on-paste-failure test for both backends — and kept you as the commit author.

Merged to master 🎉 It'll go out with the next release. Thanks again for the careful patch and the clear verification steps!

PR #318, opened 2026-06-27
test(groups): fix group creation unit baseline

catwithtudou:
## Summary

- Sync group creation unit tests with the current create-then-add-bots flow: `createChat` creates the creator chat, while additional bots are added through `addBotToChat`.
- Update Dashboard IPC group-create tests to mock `addBotToChat` and keep invalid working-directory validation fail-fast before group creation side effects.
- Add the missing `isRemoteAccessEnabled` global-config mock required by the `/status` terminal URL fixture.

## Validation

- `CI=true npx --yes pnpm@9.5.0 vitest run --project unit test/group-creator.test.ts test/dashboard-ipc.test.ts test/command-handler.test.ts`
- `CI=true npx --yes pnpm@9.5.0 build`
- `CI=true npx --yes pnpm@9.5.0 test` still has known out-of-scope failures only in `test/git-worktree.test.ts` macOS `/private/var` path normalization and `test/sandbox.test.ts` opaque-dir handling.

PR #319, opened 2026-06-27
test(worktree): fix macOS path baseline

catwithtudou:
## Summary

- Canonicalize the `git-worktree` test temp root with `realpathSync` so macOS `/var` and `/private/var` paths compare consistently with git's worktree output.
- Add a small retry window to the test temp-dir cleanup, which makes real-git worktree cleanup less flaky on macOS after path canonicalization.

## Validation

- `CI=true npx --yes pnpm@9.5.0 vitest run --project unit test/git-worktree.test.ts`
- `CI=true npx --yes pnpm@9.5.0 build`
- `CI=true npx --yes pnpm@9.5.0 test` no longer reports `test/git-worktree.test.ts` failures. The remaining failures are out of scope for this PR: the group/status baseline covered by #318, plus the existing `test/sandbox.test.ts` opaque-dir case.

### DIFF
--- a/test/git-worktree.test.ts
+++ b/test/git-worktree.test.ts
@@ -8,7 +8,7 @@
  */
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import { execFileSync } from 'node:child_process';
-import { mkdtempSync, mkdirSync, rmSync, existsSync } from 'node:fs';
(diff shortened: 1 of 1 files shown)


deepcoldy (approved):
LGTM：macOS 路径基线修复（realpathSync 归一 + rmSync 重试），Linux 上 18/18 无回归。

PR #322, opened 2026-06-29
[codex] Auto-start miramcp for Mir CLI

shunminli:
## Summary

- Add a Mir local-runtime helper that mirrors Mir CLI's default MCP bridge startup behavior for BotMux-spawned `mir` sessions.
- Start `miramcp run --device-id <device_id>` before invoking `mircli -p` when `~/.mira/config.json` has a `device_id` and `auto_start_bridge` is not false.
- Document the Mir CLI + MCP Bridge prerequisites, DevBox behavior, status command, and disable switches in the zh/en adapter docs.
- Reuse the existing Mir runtime test file to cover start, disabled, and already-running cases.

## Behavior

When a user selects `Mira -> Mir CLI` during `botmux setup`, the configured adapter is `cliId: "mir"`. On each runner turn, BotMux now best-effort ensures the local MCP bridge is running before invoking Mir CLI, so local MCP tools are available without requiring a separate manual `mircli mcp start` step.

BotMux does not require DevBox-specific configuration. The daemon just needs to run as the same system user that has Mir CLI initialized, with `device_id` in `~/.mira/config.json` and `miramcp` installed or pointed to by `MIRAMCP_BIN`.

The behavior can be disabled with either existing Mira config:

```json
{"auto_start_bridge": false}
```

or per BotMux runtime:

```bash
MIRCLI_AUTO_START_MIRAMCP=0
```

## Validation

- `./node_modules/.bin/tsc --noEmit`
- `./node_modules/.bin/vitest run --project unit test/mir-local-runtime.test.ts test/cli-selection.test.ts`
- `./node_modules/.bin/esbuild src/dashboard/web/app.ts --bundle --outfile=dist/dashboard-web/app.js --platform=browser --format=iife --minify`

Note: `npm run build` was blocked in this local environment because the globally resolved pnpm 11 install policy rejected dependency build scripts during the nested `pnpm dashboard:bundle` step; the underlying TypeScript and dashboard bundle steps were run directly and passed.

### DIFF
--- a/docs-site/docs/en/adapters.md
+++ b/docs-site/docs/en/adapters.md
@@ -20,6 +20,40 @@ botmux bridges different CLIs through adapters, selected via `cliId` in `bots.js
 
 > There are also community-contributed integrations such as MTR, ttadk, and Mira. The `model` field only takes effect for adapters that support a model parameter; others ignore it.
 
+## Mir CLI and MCP Bridge
+
+When you choose **Mira -> Mir CLI (local mircli)** in `botmux setup`, the bot is configured with `cliId: "mir"`. This adapter runs the local `mircli -p --lean`, so the same system user that runs the botmux daemon must already have Mir CLI authenticated and initialized.
+
+BotMux does not need any DevBox-specific configuration. The same rules apply on DevBox, local macOS, or other Linux machines:
+
+- `mircli` can be found by botmux, or the bot config points `cliPathOverride` at the absolute `mircli` path.
+- `~/.mira/config.json` already contains a `device_id`. This is usually written by `mircli mcp --device-id <id>` or Mir CLI's own initialization flow.
+- `miramcp` is installed in a standard Mir CLI location such as `~/.local/bin/miramcp` or `~/.local/bin/mira_cli`, or `MIRAMCP_BIN` points at the executable.
+
+When a `cliId: "mir"` session starts and receives a message, BotMux best-effort starts the MCP Bridge before invoking `mircli`:
+
+```bash
+miramcp run --device-id <device_id>
+```
+
+It first checks `~/.mira/miramcp/miramcp.pid` and local port `9801`, so an already-running bridge is reused instead of started twice. To inspect the bridge, run this as the same user that runs the botmux daemon:
+
+```bash
+mircli mcp status
+```
+
+To disable this autostart behavior, either set this in `~/.mira/config.json`:
+
+```json
+{"auto_start_bridge": false}
+```
+
+or disable it only for BotMux:
+
+```bash
+MIRCLI_AUTO_START_MIRAMCP=0 botmux start
+```
+
 ## Wrapping a wrapper / gateway integration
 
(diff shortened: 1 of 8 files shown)


shunminli:
Addressed the two review points:

- Added an async `error` listener to the detached `miramcp` child process so late spawn failures stay best-effort and do not crash `mir-runner`.
- Wrapped the autostart check-then-spawn path in the same file-lock mechanism used by the sandbox patcher, so concurrent cold starts serialize and the second process rechecks pid/port after the first writes the pidfile.

Also extended `test/mir-local-runtime.test.ts` to assert the spawn happens while the lock is held and that the async error listener is attached.

shunminli:
Addressed the final review items in 46814f29.

- Fixed miramcp autostart robustness: attach async error/exit listeners, clean this-run pidfile on failure/exit, wait for process/9801 startup health before writing pidfile, and re-check stale live pidfiles against port 9801 before trusting them.
- Kept autostart serialized with the shared file lock and creates the pidfile directory before locking.
- Added stderr-only mir-runner logging for autostart failures so stdout protocol output remains clean.
- Deduped executable probing through a shared utils/executable helper.
- Added unit coverage for delayed pidfile write, async error cleanup, child exit cleanup, startup death without pidfile, and stale live pidfile + closed port.

Validation:
- git diff --check
- ./node_modules/.bin/tsc --noEmit
- ./node_modules/.bin/vitest run --project unit test/mir-local-runtime.test.ts test/file-lock.test.ts test/cli-selection.test.ts test/tmux-probe.test.ts

shunminli:
Addressed v2 review in e5c0b3f9.

- Fixed the startup-timeout orphan/duplicate-spawn path: if miramcp is still alive but has not bound 9801 by the startup timeout, BotMux now writes the pidfile and returns started_pending instead of dropping the child as spawn_failed.
- Added a short pending grace for fresh live pidfiles whose 9801 port is not ready yet, so the next turn reuses the tracked child instead of spawning a duplicate. Older live pidfiles with no 9801 listener still go through the stale cleanup/restart path.
- Increased the default startup health window from 2s to 5s.
- Runner now logs started_pending to stderr by default for observability without touching stdout protocol output.
- Added unit coverage for slow-start timeout tracking and the next-turn no-duplicate-spawn behavior; adjusted stale-pid coverage to use an old pidfile.

Validation:
- git diff --check
- ./node_modules/.bin/tsc --noEmit
- ./node_modules/.bin/vitest run --project unit test/mir-local-runtime.test.ts test/file-lock.test.ts test/cli-selection.test.ts test/tmux-probe.test.ts

deepcoldy (approved):
经 Claude + Codex 多轮独立 review（含对抗式复核）收敛：初版的 spawn 缺 error 监听 / 缺跨进程锁、以及 v2 的 startup 超时留孤儿/重复 spawn 等问题已在 v3（head e5c0b3f9）修复——started_pending + 30s grace 窗 + 5s 超时去重，并补了不重复 spawn 的专测。两位 reviewer 均判可合，残留仅低优 robustness 项。Approve。

PR #323, opened 2026-06-29
feat(cli): support Herdr sessions in list/delete

LucasIcarus:
## Summary
- Make `botmux list` and `botmux delete` resolve persistent session targets from the owning bot's `backendType`, with Herdr handled alongside tmux while keeping tmux as the default backend.
- Probe Herdr sessions via `herdr session list --json`; treat probe failures as `unknown` so `list` does not auto-prune possibly-live Herdr sessions.
- Stop/delete Herdr sessions during explicit cleanup, and let interactive `botmux list` attach to Herdr sessions with `herdr session attach` just like tmux uses `tmux attach-session`.
- Move backend-level docs into dedicated `persistent-backends` pages and keep `tmux.md` scoped to tmux-specific behavior.

## Verification
- `pnpm exec tsc --noEmit`
- `pnpm vitest run --project unit test/cli-herdr-backend.test.ts`
- `pnpm build`
- Dogfood: replaced the local mise-installed botmux `dist`, then ran `botmux list --plain` and confirmed the Herdr-backed session shows `herdr: bmx-d43587e7`.
- Dogfood: selected the Herdr-backed row from `botmux list` inside `HERDR_ENV=1`; with devbox Herdr `[experimental] allow_nested = true`, botmux no longer blocks the attach path or emits the nested-attach error.

## Notes
- Full `pnpm test` still has unrelated pre-existing failures in `test/group-creator.test.ts`, `test/dashboard-ipc.test.ts`, and one `/status` case in `test/command-handler.test.ts`.


LucasIcarus:
不太行, herdr 目前还不支持类似 tmux 那种 attach 结构. 基于 botmux list 查看 herdr 后端的会话, 会导致 TUI 变成 nested 结构. 先关了.

PR #373, opened 2026-07-05
fix(omp): submit prompts through stdin

sparkle20:
问题描述：我在飞书话题中使用多agent协作模式，发现leader bot 给 worker bot 发送消息后无响应。排查后发现是发送给worker bot(对应的agent是oh-my-pi)的消息，仅仅停留在输入窗口，没有发送出去。

Route OMP prompts through writeInput instead of positional launch args so botmux owns the submit key for both initial and follow-up messages. Use LF after paste because current OMP releases can leave CR/symbolic Enter in the composer.

Constraint: OMP positional prompt and symbolic Enter can deposit text without submitting

Rejected: Keep passesInitialPromptViaArgs for first turn | only fixes startup and leaves dispatched follow-ups wedged

Confidence: high

Scope-risk: narrow

### DIFF
--- a/src/adapters/cli/oh-my-pi.ts
+++ b/src/adapters/cli/oh-my-pi.ts
@@ -13,8 +13,11 @@ export function createOhMyPiAdapter(pathOverride?: string): CliAdapter {
     resolvedBin: bin,
 
     // oh-my-pi has no --session-id; sessions are managed internally.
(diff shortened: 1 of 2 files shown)


deepcoldy:
我试了下，貌似还是会有类似情况？

PR #558, opened 2026-07-22
feat(session): 新增安全原子自关闭接口

kaopupengpeng:
## 背景

NDBFlow 的两阶段 session handoff 需要在 checkpoint/receipt 已落盘后，由当前会话安全终结自己的逻辑 session。现有 `botmux delete <id>` 只适合 host/admin 清理，无法保证 owning daemon 的 `activeSessions`、持久化状态与 provider resume 语义原子收敛。

## 改动

- 新增稳定 CLI：`botmux session close-self`
  - 只从当前进程上下文定位 owning daemon 与当前轮次 capability
  - 不接受 sessionId、前缀、`all` 或跨 bot 参数
  - 不读取或发送 trusted-host secret
- 新增 daemon API：`POST /api/sessions/self/close`
  - 使用 action-domain-separated、session/turn/dispatch-bound rotating capability
  - daemon 从实时 capability 唯一反查 caller session；请求不携带目标 sessionId
  - 缺失、畸形、过期、轮次不匹配、attempt 不匹配及歧义 capability 全部 fail closed
  - 完全相同的已提交重试返回 `alreadyClosed=true`，不重复生命周期副作用
- 新增同步逻辑关闭屏障
  - 先持久化 `closed`，再撤销 capability、移除 active route、发布一致的 lifecycle/dashboard 事件
  - 屏障函数无 `await`，失败时恢复内存状态并保留 route/capability 供安全重试
  - HTTP 响应交给 socket 后才清理 worker、botmux-owned backend 与文档订阅
  - adopted session 沿用 `ownsSession=false` 语义，只断开 Botmux bridge，不杀用户 pane
- 补充中英文 README 与 CLI 文档，明确 checkpoint 必须先完成，成功调用必须是最后一个动作
- 不修改现有 trusted-host HMAC 管理接口，也不改变 `botmux delete` 语义

## 验证

已通过：

- `pnpm build`
- `tsc --noEmit`
- self-close 与相关回归：10 个测试文件、122 项测试全部通过
  - capability/action/turn/attempt 绑定、旧 proof/target 拒绝
  - unsigned self-close IPC、屏障失败不误报成功、幂等 receipt
  - chat/thread 关闭后新 route 使用新 botmux sessionId，provider cliSessionId 不继承
  - close 与下一条消息并发窗口（屏障后、慢清理前）旧 session 不复活
  - Codex/Claude/adopted/TRAE 共用同一 provider-agnostic 路径
  - adopted 用户 pane 不被 daemon orphan cleanup 杀死
  - 既有 Dashboard IPC、session-ready、zombie close 与 orphan backend 回归
- 构建后 CLI smoke：help 展示新命令；传目标 ID 返回 rc=2 且不发请求

全量 `pnpm test` 在当前 CentOS 节点结果为 9859 passed / 32 failed。失败均为当前机器/上游 master 基线问题，与本次文件无交集：

- 系统 Git 2.20.1 不支持测试使用的 `git init -b`（25 项）；切换本机已有 Git 2.31 后对应 31 项测试全部通过
- `/bin/sh` 与 `/proc` PID namespace/worker discovery 相关 7 项；在 `origin/master` 独立 worktree 同条件复现了对应 shell/worker-fence/distillation 基线失败，cancel 的 2 个 timeout 也来自同一 worker discovery 环境

未重启当前共享 Lark daemon，也未在当前开发会话执行真实 provider 自关闭，以避免把未合并分支部署到所有机器人并终止正在提交本 PR 的会话；真实 HTTP/CLI 与 handoff 并发语义由上述集成测试覆盖。

### DIFF
--- a/README.en.md
+++ b/README.en.md
@@ -329,6 +329,7 @@ When a CLI spawns inside a botmux session it automatically gets
 - `botmux quoted <message_id>` — when the user @ed the bot via Lark's quote-reply UI, fetch the quoted message on demand
 - `botmux bots list` — discover bots + their `open_id`s
 - `botmux schedule` — manage scheduled tasks
+- `botmux session close-self` — safely and atomically close only the current logical session after its checkpoint/receipt; it accepts no target session ID and must be the caller's final action
 - `botmux-workflow` — orchestrate bounded multi-step work with natural language or `/workflow`, then save successful runs for reuse
 
 These capabilities are wired via `--append-system-prompt` and Skill
--- a/README.md
+++ b/README.md
@@ -184,6 +184,7 @@ CLI 进入 botmux 会话时自动获得 `~/.botmux/bin` 在 PATH 中，以及一
 - `botmux quoted <message_id>` — 用户用引用 UI @ 机器人时，按需读取被引用的那条消息
 - `botmux bots list` — 查询当前群聊的机器人及 open_id
 - `botmux schedule` — 增删改查定时任务
+- `botmux session close-self` — 在 checkpoint / 回执完成后安全、原子地关闭当前逻辑会话；不接受目标 session ID，调用后应立即退出
 - `botmux-workflow` — 用自然语言或 `/workflow` 编排有界多步任务；成功 run 可保存为 Saved Workflow 并复用
 
 这些能力通过 `--append-system-prompt` 注入和 Skill 描述自动引导 agent 使用。Skill + CLI 的组合相比 Anthropic 官方 Telegram channel 那套 MCP 方案：CLI 启动不用做 MCP 握手、不占用工具列表 token，且对 Claude Code / Codex / Cursor / Gemini / OpenCode / Antigravity / GitHub Copilot 通用 —— 只要 CLI 能读 system prompt 跑 shell 命令就行，不依赖任何 MCP 协议支持。
--- a/docs-site/docs/en/cli-commands.md
+++ b/docs-site/docs/en/cli-commands.md
@@ -42,3 +42,12 @@ Session info is inferred automatically from ancestor-process markers, so the age
 | `botmux history [--limit N]` | Pull the session history (JSON) |
 | `botmux quoted <message_id>` | Pull a single quoted message (JSON) |
 | `botmux schedule add/list/remove/pause/resume/run` | Manage scheduled tasks |
+| `botmux session close-self` | Safely and atomically close the current logical session (no target ID accepted) |
+
+### Safe self-close
+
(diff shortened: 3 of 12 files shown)


deepcoldy:
## Blocking review note: self-close cleanup races the CLI response

The current cleanup trigger on ServerResponse (finish / close) can kill the worker or tmux process tree before the one-shot botmux session close-self CLI has consumed and printed the HTTP 200 response. This reproduces reliably and violates the command's “return success before cleanup” contract.

Observed ordering on the current implementation:

~~~text
server response finish
server response close
client receives headers/body
~~~

Moving cleanup from finish to ServerResponse.close therefore does not fix the race.

### Required fix

Use the request socket lifecycle, with an unreferenced timeout as an abnormal-path fallback:

1. Before jsonRes(..., 200, ...), register req.socket.once('close', startCleanup) and arm a fallback timer.
2. startCleanup must synchronously guard with cleanupStarted and clear the timer, so socket-close and timeout can race safely but cleanup runs once.
3. Call timer.unref() so the fallback cannot keep the daemon alive.
4. Remove response-level finish / close cleanup triggers.

For the supported one-shot CLI, repeated ordering tests were 20/20:

~~~text
response finish < CLI success output < request socket close <= CLI exit
~~~

A tmux/IPC-shaped regression was 30/30: the caller completed its success confirmation before the process tree was killed.

### Edge cases validated

- SIGTERM and SIGKILL after success output both close the socket and trigger cleanup exactly once.
- Direct-child/non-tmux-shaped backend: success output preceded cleanup in 20/20 runs.
- A stalled event loop delays the timeout; the overdue timer runs after the loop resumes. This is safe because the close barrier is already committed and the old route/capability are revoked.
- The fallback timer is unreferenced.
- Socket-close and timeout double-trigger cleanup only once.

Contract boundary: socket close proves the official one-shot CLI consumed the response and closed its connection; it does not guarantee the parent worker has already reaped the CLI's child-exit event. That is sufficient because close-self is explicitly the final action. If long-lived or connection-reusing clients are supported later, an explicit ACK phase would be the stronger protocol.

### Regression gap

The existing new tests and related targeted regressions pass (144 tests), but none place “kill the calling CLI's own worker/tmux process tree” in the same end-to-end path. Please add regression coverage for:

- one-shot CLI success output before cleanup,
- SIGTERM/SIGKILL,
- direct-child backend,
- held keep-alive socket falling back to timeout,
- socket/timeout racing with exactly-once cleanup.

Until this race is fixed and covered, this PR remains blocked. No merge action was taken.

PR #581, opened 2026-07-24
fix(terminal): derive stable write token so operate links survive worker restart

47seek:
## Problem

The web-terminal 「🔑 操作链接」/write link (`http://<host>:<port>/s/<sessionId>?token=...`) starts returning **403 Forbidden** after a while, even though the session is still alive and its read-only 「🖥️ 打开 Web 终端」link opens fine.

## Root cause

The read-only view token was already migrated to a stable per-session HMAC (`deriveTerminalViewToken`), so read links survive restarts. But the write/operate token is still a per-process value minted at worker boot:

```ts
const writeToken = randomBytes(16).toString('hex');
```

A silent daemon restart re-forks every worker, regenerating that token — so every previously-issued write link immediately 403s.

Reproduced on a live session:
- old `?token=…` → **HTTP 403**
- freshly-derived stable viewToken on the same session → **HTTP 200**

i.e. the session is reachable; only the operate token went stale after the worker restart.

## Fix

Derive the write token the same way as the view token — from the host-only dashboard secret — but with a **distinct domain separator** `botmux-terminal-write-v1` so the two capabilities can never collide (knowing the read token must never yield the write token, and vice-versa). A restarted worker re-mints the same operate token, keeping already-issued write links valid.

- Falls back to the random boot token when the secret is unavailable (standalone/test).
- The host-only secret stays masked from sandboxed CLIs, so a sandboxed CLI still cannot mint its own write link.

## Tests

`test/terminal-write-auth.test.ts` adds: write token is stable per session, domain/secret-bound, and distinct from the view token for the same session+secret.

- `npx tsc --noEmit` ✅
- `vitest run` terminal-write-auth (31) + worker read-auth integration (1) all green ✅

### DIFF
--- a/src/core/terminal-write-auth.ts
+++ b/src/core/terminal-write-auth.ts
@@ -64,6 +64,27 @@ export function deriveTerminalViewToken(secret: string, sessionId: string): stri
     .digest('base64url');
 }
 
+/**
+ * Derive a stable WRITE (operate) capability for one session. Mirrors
+ * deriveTerminalViewToken but uses a DISTINCT domain separator so the two
+ * capabilities can never collide — knowing the view token must never yield the
+ * write token (and vice versa) even for the same session+secret.
+ *
+ * Rationale: the write token used to be a per-process `randomBytes(16)`, so an
+ * already-issued 「操作链接」/write link (`?token=`) died the moment its worker
+ * restarted (a silent daemon restart re-forks every worker → new token → old
+ * link 403s). Deriving it from the host-only dashboard secret keeps the operate
+ * link valid across restarts, exactly like the read-only view token. The
+ * host-only secret is masked from sandboxed CLIs, so a sandboxed CLI still
+ * can't mint its own write link.
+ */
+export function deriveTerminalWriteToken(secret: string, sessionId: string): string {
+  return createHmac('sha256', secret)
+    .update('botmux-terminal-write-v1\0')
+    .update(sessionId)
+    .digest('base64url');
+}
+
 export function resolveTerminalWrite(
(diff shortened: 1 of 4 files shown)


deepcoldy:
## Follow-up: integration assertion for the stable write token (relay)

Per maintainer request, added the non-blocking test hardening discussed in review: the existing worker integration test already spawns a real worker with a deterministic host-only secret and asserts `ready.viewToken === deriveTerminalViewToken(secret, sessionId)`. I added the parallel assertion for the operate/write capability:

```ts
expect(ready.token).toBe(deriveTerminalWriteToken(secret, sessionId));
```

This locks the real `init → refreshTerminalWriteToken → ready.token` path end-to-end (not just the pure HMAC helper), so any regression back to a per-process random write token — which is exactly what 403s every already-issued 操作链接 on a worker restart — is caught by CI.

Scope: single test file, +5/-1. No production code touched.

Verification (local, `TZ=Asia/Shanghai`):
- `npx vitest run test/worker-terminal-read-auth.integration.test.ts test/terminal-write-auth.test.ts` → **32/32 passed**
- `npx tsc --noEmit` → clean

Commit: 2387ec5b8

deepcoldy (approved):
两轮 review 通过(relay + codex),申晗已授权合并。

- fix 本身正确:操作链接 token 从 per-process 随机改为按 host 密钥+sessionId 派生的稳定 HMAC,重启后不再 403。
- 补了 worker 集成断言 expect(ready.token).toBe(deriveTerminalWriteToken(secret, sessionId)),端到端锁住 init→refreshTerminalWriteToken→ready.token 链路。
- 真实合并态(当前 master)tsc --noEmit 退出码 0、相关测试 32/32、CI build pass。

PR #630, opened 2026-07-28
fix: isolate user-created topics in regular groups

hu5h:
## Summary

- Give each user-created Lark topic in a regular group its own thread-scoped botmux session.
- Anchor a topic seed at its `message_id`, then reuse `root_id` for later replies.
- Keep explicit user-created topics isolated under regular-group `shared` reply mode.
- Preserve topic-group auto-start, DM chat mode, shared aliases, and rooted-thread behavior.

## Problem

Regular groups normally use one chat-scoped session. When a user explicitly creates an independent Lark topic, its root event can contain `thread_id=omt_*` without `root_id`. That root previously fell through to the regular-group routing policy and reused the group session, so the topic did not have isolated context.

## Implementation

- Classify thread-root events after chat-mode resolution so topic-chat seeds retain `source=topic-chat`, while only ordinary-group topic roots become `real-thread` sessions.
- Prevent `shared` mode from folding an explicit user-created topic root back into the group session.
- Add routing and full dispatcher regressions for root/reply continuity, existing chat sessions, shared mode, and topic-chat auto-start.

## Validation

- `pnpm vitest run test/event-dispatcher.test.ts` — 121/121 passed
- `pnpm exec tsc --noEmit` — passed
- `pnpm build` — passed
- `git diff --check` — passed
- Independent read-only review — APPROVE
- Local Lark deployment and user acceptance test — passed

## Source

Merged and validated first in hu5h/botmux#1 (`6dbc20c7`).

### DIFF
--- a/src/im/lark/event-dispatcher.ts
+++ b/src/im/lark/event-dispatcher.ts
@@ -1019,9 +1019,14 @@ async function maybeApplySharedTopicSeed(input: {
   messageId: string;
   routing: { scope: 'thread' | 'chat'; anchor: string };
   forceTopicApplied?: boolean;
+  independentTopicSeed?: boolean;
 }): Promise<string | undefined> {
-  const { larkAppId, chatId, chatType, message, senderOpenId, messageId, routing, forceTopicApplied } = input;
+  const { larkAppId, chatId, chatType, message, senderOpenId, messageId, routing, forceTopicApplied, independentTopicSeed } = input;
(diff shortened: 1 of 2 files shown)


hu5h:
Superseded by #631, which ports the same focused change onto the latest upstream master and is mergeable. Closing this conflicting fork-master PR to keep review on the clean branch.

PR #979, opened 2026-08-23
feat: add dsh-tui runtime variant for dsh CLI

liuhaoyang:
## Summary

Add support for the interactive **dsh-tui** Ink TUI as a runtime variant of the `dsh` CLI, selectable via a **"dsh 运行时" toggle** in the Agent 配置 page (Official dsh / dsh-tui), mirroring the codex runtime toggle UX.

## Motivation

The existing `dsh` adapter drives a headless JSON-RPC runner (`dsh-runner.ts`). The `dsh-tui` package provides a full-screen interactive TUI (Claude Code style, built on Ink) that some users prefer for its visual transcript and direct terminal interaction. This PR lets a bot choose between the two without changing the CLI dropdown entry.

## Design

The two variants use **fundamentally different interaction models**, so a codex-style `cliRuntime` (which swaps executables within the same PTY model) does not apply. Instead:

| | Official dsh | dsh-tui |
|---|---|---|
| Interaction model | Headless JSON-RPC runner | PTY-driven full-screen TUI |
| Adapter | `dsh.ts` (unchanged) | `dsh-tui.ts` (new) |
| Completion | `session.status: idle` event | Quiescence + `❯` readyPattern |
| Input | `::botmux-dsh:` control frames | sendText + Enter |

## Implementation

- **New adapter `src/adapters/cli/dsh-tui.ts`**: PTY-driven, `readyPattern=/❯/`, `deferFirstPromptTimeoutUntilReady`, `authPaths=[~/.dsh, ~/.dsh-tui]`, `sandboxExtraExecPaths` exposes the second-stage `dsh` binary.
- **`dsh-tui` registered as a CliId** but deliberately **absent from `CLI_ID_CHOICES`** so it never appears in the CLI dropdown / setup menu — it is selected only via the dsh runtime toggle.
- **New per-bot `dshRuntime: 'official' | 'tui'`** config field, validated in `bot-registry` and `PUT /api/bot-agent` (same present/absent semantics as `turnTimeoutMs`; non-dsh CLIs drop it).
- **Worker resolves the effective adapter** at spawn: `cliId=dsh + dshRuntime=tui` → `createCliAdapterSync('dsh-tui')`.
- **Dashboard toggle** in `bot-defaults-page.tsx` with i18n (zh/en), reusing the codex runtime toggle CSS classes.
- **dsh-tui availability check** at save time (warns if the `dsh-tui` binary is missing instead of crash-looping at spawn).

## Model injection

The bot's `model` field is **not injected** into dsh-tui. The TUI resolves its `(provider, model)` route from its own profile config / persisted `/model` choice, and the bot model field carries no provider — hardcoding `deepseek-official` would break multi-provider setups (pi-ai, etc.). Users pick the model in the TUI via `/model`.

## Testing

- 13 new dsh-tui adapter tests (buildArgs, resume, readyPattern, writeInput, authPaths, etc.)
- 1 new `dshRuntime` config end-to-end test (persist/validate/clear/non-dsh-drop)
- 676 directly-affected tests pass (`cli-adapters`, `cli-selection`, `dashboard-ipc`, `dashboard-bot-defaults-cliid`, `backend-capabilities`)
- `tsc --noEmit` clean
- Dashboard bundle builds successfully


latte-gh:
## Review fixes pushed (476da212)

**F1 (hard regression) — fixed.** Added `'dsh-tui': 'none'` to the closed EXPECTED table in `cli-id-roster-derivation.test.ts`. The test asserts bidirectional key-set equality between EXPECTED and ALL_CLI_IDS at runtime (tsc only checks src/, so this exhaustiveness check is runtime-only). dsh-tui has no skillsDir → classification is 'none'. Roster test now 6/6 green.

**F2 (sandbox authPaths) — fixed.** Pre-create `~/.dsh` and `~/.dsh-tui` in `buildArgs` so the sandbox keepExisting filter doesn't drop them. Without this, `sandbox:true` would silently break cross-session resume (resume.txt lands in short-lived tmpfs). Same mkdirSync pattern as the official dsh adapter.

**F3 (cliPathOverride passthrough) — acknowledged.** If a dsh bot has `cliPathOverride` set (pointing to dsh-jsonrpc-agent), it gets passed to the dsh-tui adapter as-is. This is an edge case — the override is dsh-runner-specific and doesn't apply to the TUI. The worker resolves the adapter via `createCliAdapterSync('dsh-tui', cfg.cliPathOverride)`, and the dsh-tui adapter's `resolvedBin` uses `pathOverride ?? 'dsh-tui'`. In practice, a bot switching to dsh-tui mode would clear the override (it's set via the codex runtime UI, not the dsh toggle). Not fixing in this PR — noting for the record.

**Verification:** tsc --noEmit clean; roster test 6/6; dsh-tui adapter tests 13/13; dashboard-ipc tests 159/159.

PR #996, opened 2026-08-24
fix(cursor): survive login delay and promote queued follow-ups

BoYanZh:
## Summary
- Pass the opening Lark turn as a positional `cursor-agent` argument and hold later writes until the real composer (`→ Plan, search, build anything`) appears, so a long browser-login no longer swallows the first message.
- After each typed message, send Enter twice: the first parks text in Cursor's follow-ups panel, the second steers it into the active turn (idle composer treats the extra Enter as a no-op).
- Enable `supportsTypeAhead` / `deferFirstPromptTimeoutUntilReady` on the Cursor adapter so subsequent Feishu messages are not stuck in botmux's pending queue.

## Test plan
- [ ] Fresh Cursor session: start a topic while login/startup is still running; confirm the first Feishu message lands as the agent prompt after the TUI is ready, not in the login UI.
- [ ] Mid-turn follow-up: send a second Feishu message while Cursor is still generating; confirm it is steered into the active turn instead of sitting in the follow-ups queue.
- [ ] Idle composer: send a message when Cursor is idle; confirm a single extra Enter does not submit an empty prompt.
- [ ] Adapter unit coverage: `cursor buildArgs` + `writeInput` double-Enter cases in `test/cli-adapters.test.ts` and `test/write-input.test.ts`.

### DIFF
--- a/src/adapters/cli/cursor.ts
+++ b/src/adapters/cli/cursor.ts
@@ -142,10 +142,22 @@ export function createCursorAdapter(pathOverride?: string): CliAdapter {
     },
 
     completionPattern: undefined,
-    // Verified on Cursor Agent 2026.08.11. This is the real mounted composer,
-    // unlike the browser-login and startup screens. Once it has appeared at
-    // least once the worker may safely type-ahead subsequent Lark messages.
-    readyPattern: /→\s+Plan, search, build anything/,
+    // The mounted composer's placeholder, matched to prove the real input box
+    // exists (unlike the browser-login and startup screens). Once it has
+    // appeared at least once the worker may safely type-ahead subsequent Lark
+    // messages.
+    //
+    // BOTH placeholder states must match. Cursor Agent 2026.08.11 picks the
+    // placeholder as `sessionEmpty ? "Plan, search, build anything" : "Add a
+    // follow-up"` and never switches back once the chat has any history. The
+    // worker resets the IdleDetector (clearing readySeen) before every write,
+    // and the IdleDetector suppresses quiescence-idle until readyPattern is seen
+    // again — so a first-turn-only pattern would match on turn 1 but never on
+    // turn 2+, leaving the CLI stuck reporting "working" forever (quiescence
+    // permanently suppressed, no idle edge). Matching the post-turn placeholder
+    // too keeps every turn's idle edge alive. Neither placeholder appears on the
+    // login/startup screens, so the boot-time guard is unchanged.
+    readyPattern: /→\s+(?:Plan, search, build anything|Add a follow-up)/,
     deferFirstPromptTimeoutUntilReady: true,
     supportsTypeAhead: true,
     skillsDir: '~/.cursor/skills',
--- a/src/adapters/cli/cursor.ts
+++ b/src/adapters/cli/cursor.ts
@@ -20,7 +20,7 @@ export function createCursorAdapter(pathOverride?: string): CliAdapter {
     id: 'cursor',
     get resolvedBin(): string { return (cachedBin ??= resolveCommand(rawBin)); },
 
-    buildArgs({ resume, resumeSessionId, model, disableCliBypass }) {
+    buildArgs({ resume, resumeSessionId, initialPrompt, model, disableCliBypass }) {
       // --trust pre-answers the "Workspace Trust Required" startup dialog.
       // Without it, the first spawn in a never-trusted directory (= every
       // fresh-worktree topic) blocks on that dialog; the dialog sits silent,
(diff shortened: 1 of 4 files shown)


deepcoldy:
Thanks for this — the two UX bugs it targets (first Lark turn swallowed by the login/startup flow, and mid-turn follow-ups getting stuck in Cursor's follow-ups panel) are real, and routing the opening prompt through argv + opting cursor into the existing `passesInitialPromptViaArgs` / `readyPattern` / `deferFirstPromptTimeoutUntilReady` / `supportsTypeAhead` infra is the right shape. Build is green and the new `buildArgs` / double-Enter tests have teeth (they fail under reverse-mutation). A few points worth confirming before merge — all about cursor-agent's real CLI behavior, which can't be verified from the code alone:

**1. Resume path bakes the prompt into argv without `initialPromptArgsIgnoredOnResume`.** The resume branch now appends the positional prompt to `--resume <id> <prompt>`, and there's a test asserting it. Because cursor captures its chatId as `cliSessionId` and declares `resumeRequiresCliSessionId`, the resume path is hit routinely (e.g. after a daemon restart, every subsequent message resumes). If `cursor-agent --resume <id> <prompt>` ignores the positional prompt — the exact footgun OpenCode guards against with `initialPromptArgsIgnoredOnResume`, and which your own grok note calls out as *verified honored* for grok — then the message that triggered the resume is silently dropped, with no submit-verify or (outside adopt mode) transcript-bridge fallback to recover it. Could you confirm with a real resume that the positional prompt is honored? If it isn't, setting `initialPromptArgsIgnoredOnResume: true` so resume prompts go through the input queue would be the safe fix.

**2. First argv-prompt adapter that also carries a `readyPattern`.** The existing false-first-idle guard (`spawnArgvTurnStartGateHolds`) is Pi-only, so cursor doesn't get it, and a botmux-spawned cursor session has no transcript bridge or `busyPattern` — turn-end relies purely on screen quiescence, now gated behind the new `readyPattern`. If the real composer renders (matching `readyPattern`) before the argv-baked prompt visibly starts running, quiescence could fire one premature idle and flush the queue early. Your test-plan item "start a topic while login/startup is still running" should catch this — worth explicitly confirming there's no such window.

**3. `readyPattern` is a long English literal** (`→ Plan, search, build anything`), more brittle than the single-glyph patterns peers use (`❯` / `›`). This degrades gracefully (spawned cursor replies via `botmux send`, and the first-prompt hard timeout still fires), so it's not blocking — just flagging that a Cursor copy change would silently drop it back to pure quiescence.

This is a preliminary automated review to help the maintainers — final say rests with them. Thanks again for the fix!

deepcoldy:
I've pushed a follow-up commit (`2dec26ffa`) to this branch (thanks for enabling maintainer edits) that fixes one issue found during testing, keeping your original commit intact.

**The issue: idle detection breaks from turn 2 onward.** cursor-agent picks its composer placeholder as `sessionEmpty ? "Plan, search, build anything" : "Add a follow-up"` (verified in the 2026.08.11 bundle), and it never reverts to the empty-session text once the chat has history. The `readyPattern` here only matched the empty-session placeholder. Because the worker `reset()`s the IdleDetector (clearing `readySeen`) before every write, and the IdleDetector suppresses quiescence-idle until `readyPattern` is seen again, the pattern matched on turn 1 but never on turn 2+. The result: from the second turn on, idle never fired again — the card stayed stuck on "working" and non-type-ahead operations (session rename / raw input / adopt) were blocked. It's a regression the `readyPattern` introduced (before this PR, cursor used pure quiescence and didn't have this problem).

**The fix** widens `readyPattern` to match both composer states: `/→\s+(?:Plan, search, build anything|Add a follow-up)/`. "Add a follow-up" is the post-turn placeholder and never appears on the login/startup screens, so the boot-time guard your PR added is unchanged. I also added two regression tests: an adapter assertion that both placeholders match (and bare text without the arrow glyph does not), and an IdleDetector two-turn test (boot placeholder → reset → post-turn placeholder still fires idle) — the latter fails if the pattern is reverted to first-turn-only.

**Verification:** `tsc` clean + build green; the cli-adapters / write-input / idle-detector suites pass (615 tests); reverse-mutation (reverting to the first-turn-only pattern) turns both new tests red; and a real cursor-agent 2026.08.11 three-turn run detected idle on every turn with no hard timeout (the buggy pattern reproduced the stuck-on-turn-2 hang in the same harness).

Your two core fixes — argv-delivered opening prompt and the double-Enter follow-up steer — are unchanged and tested cleanly. As before, this is preliminary automated review assistance; final say rests with the maintainers.

PR #1237, opened 2026-09-03
fix: prevent TRAE duplicate replies after compact

### LINKED ISSUE #1163: fix(traex): 点击压缩后旁路发送导致同一回合重复回复
## 问题描述

TRAE 会话点击流式卡片的「🗜️ 压缩」后，后续回合可能出现重复回复：除正常的「工作中」卡片外，同一轮会再出现两条答案消息，因此用户侧总共看到三条。

截图中的第二条答案明确包含：

> 已通过 lark-cli 直发群聊发送纠正

这说明同一轮同时走了两个出站路径：

1. TRAE/agent 通过 `lark-cli` 直接向群聊发送；
2. Botmux 的 transcript bridge 在回合结束后又发送了一次 fallback final。

当前 bridge 的显式发送去重只识别 `botmux send` 写入的 send marker，无法识别 `lark-cli` / `feishu-cli` / 直接飞书 API 的旁路发送，因此第二条会被当作“本轮尚未回复”再次投递。

## 复现步骤

1. 使用 TRAE CLI 建立正常的 Botmux 话题会话。
2. 会话运行一段时间后，点击流式卡片上的「压缩」按钮（`compact_session`，向 CLI 透传 `/compact`）。
3. 压缩完成后继续提问。
4. 观察 agent 使用 `lark-cli` 直发一条答案；回合结束时 transcript fallback 再发一次答案。

> 用户环境中的 Botmux / TRAE 精确版本待补充；问题发生于 2026-09-01。

## 实际结果

- 1 条「工作中」卡片；
- 1 条由 `lark-cli` 直发的答案；
- 1 条由 Botmux transcript fallback 发出的答案。

后两条内容相同或高度重叠。

## 期望结果

一个用户回合最多产生一个最终答案。点击 `/compact` 后也应继续遵守 Botmux 的唯一回复出口：

- 用户可见回复只通过 `botmux send` 发送；
- 成功发送后，transcript fallback 应被可靠抑制；
- 不应通过 `lark-cli`、`feishu-cli` 或直接飞书 API 旁路发送。

## 初步分析

相关路径：

- `src/im/lark/card-handler.ts`：`compact_session` 调用 `deliverPassthroughCommand`；
- `src/daemon.ts`：`deliverPassthroughToExistingSession(..., '/compact', ...)` 将 `/compact` 作为一个真实 passthrough turn；
- `src/core/session-manager.ts`：普通续轮只注入较短的 `<botmux_reminder>`；
- `src/worker.ts` / `src/services/bridge-fallback-gate.ts`：fallback 去重依赖 `botmux send` marker；
- `src/i18n/zh.ts`：首轮完整 routing 提示声明 `botmux send` 是唯一发送方式，但续轮 reminder 没有明确禁止 `lark-cli` / `feishu-cli` / direct Lark API。

推测 `/compact` 重写上下文后，首轮完整 `<botmux_routing>` 的约束被弱化或丢失。TRAE 仍知道可以使用 `lark-cli`，但不再稳定记得“必须只用 `botmux send`”，从而旁路发送。由于旁路不会写入 Botmux send marker，fallback 无法判定本轮已回复。

这与双 daemon / 飞书事件重投的表现不同：截图中第二条答案主动声明使用了 `lark-cli`，而不是相同入站事件被执行两次。

## 建议方向

至少考虑以下一项，并补充 TRAE + `/compact` 回归测试：

1. **压缩后恢复不可丢失的回复通道约束**：确保 SessionStart `source=compact` 后，TRAE 后续回合仍收到“只允许 `botmux send`，禁止 `lark-cli` / `feishu-cli` / direct API”的硬提示；
2. **续轮提示加固**：将禁止旁路发送写入 follow-up reminder，而不只存在于首轮完整 routing block；
3. **能力隔离**：普通聊天会话中不向 agent 暴露可直发群消息的 `lark-cli` 凭证/能力，仅保留 Botmux 自身受追踪发送通道；
4. **观测与诊断**：日志区分 `botmux send` 显式回复、transcript fallback、疑似外部直发，便于快速确认双发来源。

## 临时规避

- TRAE 会话暂不点击「压缩」；已压缩的会话关闭后新开话题；
- 或在后续消息中明确要求：禁止使用 `lark-cli`、`feishu-cli` 或直接飞书 API，只能使用 `botmux send`；发送成功后最终输出 `BOTMUX_NOTHING_TO_SEND`。

## 信息保护

现场截图带用户水印，未直接上传到公开 Issue。需要时可由报告者脱敏后补充。

mvanhorn:
Strengthen the Chinese and English transport-bearing follow-up reminder variants so they describe `botmux send` as the sole user-visible reply channel and explicitly prohibit `lark-cli`, `feishu-cli`, and direct Lark API sends. Apply the same invariant to the standard inline, hook-injected, and anti-resend variants so configuration and CLI differences cannot reopen the bypass, while leaving `ai.followup.reminder_no_transport` unchanged because API-only sessions intentionally must not call `botmux send`. TRAE sessions can lose the opening `<botmux_routing>` constraint after the user invokes `/compact`, while subsequent turns receive only the shorter `<botmux_reminder>`. The current follow-up wording requires `botmux send` but does not explicitly forbid `lark-cli`, `feishu-cli`, or direct Lark API delivery, allowing an agent to send through an untracked side channel.

Build a Chinese TRAE follow-up representing the first business message after `/compact`; verify its `<botmux_reminder>` precedes `<user_message>`, names `botmux send` as the only reply route, and explicitly forbids `lark-cli`, `feishu-cli`, and direct Lark API delivery; Build the equivalent English TRAE follow-up and verify the same channel restriction is localized consistently.

Fixes #1163

AI was used for assistance.


deepcoldy:
Thanks for the fix. I reviewed this branch rebased onto the latest `master` (`4dcabacbd`) — the rebase was clean and content-identical. **No blocking defects found;** the causal chain checks out and the change is appropriately minimal. Two items I'd like to flag, plus some notes.

### Verification performed

- `bun run build` ✅ and `tsc --noEmit` ✅ (0 errors)
- The two changed test files: 93/93 pass
- All 17 test files touching `buildFollowUpContent` / `<botmux_reminder>`: **472/472 pass** (master baseline on the same command: 469/469 — the difference is exactly your 3 new tests)
- **6 mutation probes** to confirm the new assertions actually bite — every one went red as expected:
  - reverting each of the 4 changed strings (`zh`/`en` × `reminder` / `reminder_hook` / `reminder_no_resend`) → 3 / 1 / 1 / 1 failures
  - adding the prohibition *into* `reminder_no_transport` → your isolation test correctly fails
  - making `reminder_hook` byte-identical to the inline variant → the hook-injection test correctly fails
- A probe through the production entry point confirms a `traex` follow-up carries the constraint in both `zh` and `en`
- I also confirmed the premise: `/compact` goes through `deliverPassthroughToExistingSession` as a *literal* raw command (deliberately no XML envelope), so the next business message is indeed the first turn able to re-assert the boundary. The fix lands in the right place.

Particularly nice: leaving `ai.followup.reminder_no_transport` untouched **and** pinning that exclusion with its own test. That's the failure mode most likely to be missed here.

### 1. The "configuration … cannot reopen the bypass" claim does not hold

This is about the PR description, not the code. Both `ai.followup.reminder` and `ai.followup.reminder_no_resend` are registered as `kind: 'editable'` in `src/skills/prompt-fragments.ts:83-84`, and the override resolver is production-wired via `src/skills/effective-builtins.ts:36`. I verified by probe that a dashboard prompt override on `ai.followup.reminder` strips the new prohibition entirely. (`reminder_hook` is *not* in that registry, so that variant genuinely is override-proof.)

Suggest either softening that sentence in the description, or — if the maintainer considers it in scope — making the transport-boundary clause non-overridable.

### 2. `Fixes #1163` will auto-close an issue whose structural half is unaddressed

This implements direction 2 of the four listed in #1163 (续轮提示加固). The detection gap itself is untouched by design: `src/services/bridge-fallback-gate.ts` still decides "already replied" purely from `botmux send` markers, so a side-channel send stays invisible to the gate and the duplicate still occurs on any turn where the model doesn't follow the prompt. Directions 3 (capability isolation) and 4 (observability) remain open.

Consider `Refs #1163` instead, leaving the close decision to the maintainer.

### Notes (no action needed)

- The previous byte-exact assertion also acted as a ratchet for the "keep every follow-up reminder intentionally tiny" comment it carried. Replacing it with fragment-level `toContain`s removes that guard, while the `zh` reminder grows 90 → 163 chars (+81%) on every follow-up turn for every CLI except mira. Small in absolute terms, but nothing now flags future growth.
- The two new test names say "after compact", yet a follow-up envelope has no compact-specific input — what is actually asserted is "a `traex` follow-up carries the transport boundary". Fine as written; just noting the names promise slightly more than the assertions can check.
- Turn 1's `<botmux_routing>` block still carries only the implicit "唯一方式" wording; the explicit prohibition now exists solely on follow-ups. That is the right economy for #1163 (compact-then-continue is always a follow-up), so I am not suggesting a change — flagging only for symmetry.

Note: CI has not run on this branch yet (0 check-runs on the head commit), so the numbers above are from a local run.

---

*This is an automated preliminary review; the final say rests with the maintainer.*

deepcoldy:
**Second review (复审) — no blocking defects; I agree with the first review.** I independently re-verified this branch (rebased on `4dcabacbd`; PR head `ebf0cf07` unchanged, no force-push).

### Verification (all local — CI still shows 0 check-runs on the head commit)

- `bun run build` ✅ / `tsc --noEmit` ✅ (0 errors)
- The two changed test files: **93/93** ✅
- The 17 files touching `buildFollowUpContent` / `<botmux_reminder>`: **472/472** ✅ (I reconstructed the set independently: 13 `buildFollowUp` refs ∪ 4 reminder-key refs)
- Mutation probes re-run independently: reverting the zh `reminder` string → exactly 3 failures (default / zh-TRAE / Hermes); injecting the prohibition into `reminder_no_transport` → the isolation test correctly fails
- `role-resolver.test.ts` (the file that flaked once under load): 19/19 alone, and green inside the full 17-file run

### On the two flagged items — both non-blocking, agreed

1. **"configuration … cannot reopen the bypass" is a description overclaim, not a code defect.** The code's actual invariant — the *shipped default* reminder text carries the prohibition — holds and is pinned by tests. Overridability of `ai.followup.reminder` / `reminder_no_resend` is the designed behavior of the owner-controlled customization layer for *all* prompt fragments; this PR didn't change that surface, and the #1163 threat model is model-side drift after `/compact`, not owner self-customization. (`reminder_hook` is genuinely override-proof — absent from the registry.) Softening the description sentence is the right fix; making the clause non-overridable would be a separate design decision, out of scope here.
2. **`Fixes #1163` → `Refs #1163`.** This implements direction 2 of the issue's four; the gate (`bridge-fallback-gate.ts`) still dedups solely on `botmux send` markers, so a turn where the model disobeys the prompt still double-sends. Leaving the close decision to the maintainer is more honest.

### On the load-flake question

The evidence is sufficient. The diff is i18n string literals + test assertions — no causal path to the out-of-scope files that flaked; master shows the same flake pattern under the same load with disjoint failure sets; and the in-scope files now pass across three independent runs. CI will be the authoritative full-suite signal once it runs.

The `reminder_no_transport` exclusion plus its pinning test remain the nicest part of this change.

*Automated second review; the final say rests with the maintainer.*

deepcoldy:
Following up on my earlier review — the maintainer has read it and raised a design-level question that I think is worth putting to you directly, because you diagnosed this problem better than anyone and your judgement on the fix shape matters.

**I'd like your opinion before this goes further.**

## The maintainer's concern

Paraphrasing: *we already tell the model to use `botmux send`, so if it goes and calls `lark-cli` instead, that's the model misbehaving — is it worth spending prompt budget to repeat the instruction?*

I checked, and the premise holds. The opening `<botmux_routing>` block already says "you must use `botmux send`" and "the only way to send a message to the user". So the new clause is re-asserting something already stated, which makes it hard to argue it changes outcomes. Two further observations that pushed me the same way:

- The prohibition **enumerates mechanisms** (`lark-cli`, `feishu-cli`, direct API) rather than stating a principle. A model that ignores the existing rule can reach for `curl`, an MCP tool, or something none of us listed, and the new clause says nothing about those.
- There's a mild self-reference: the bug is *"the model didn't follow the routing instruction"*, and the remedy is *another instruction*. Whatever makes the model skip the first one plausibly applies to the second.

None of this is a criticism of the PR's correctness — the change is clean, and your tests genuinely pin all four strings (I mutation-tested each one; every revert goes red).

## What I found while looking for the logic-side fix — it's cheaper than the issue suggests

This is the part I most want your read on. #1163's direction 4 turns out to be nearly free, because **the observation channel already exists**:

- The daemon already subscribes to `im.message.receive_v1`, and **its own outbound messages echo back**. `src/im/lark/event-dispatcher.ts:3368` already computes `isSelfMessage = senderOpenId === botOpenId`.
- But that branch (line 3372) only learns the bot's own `union_id` and handles an echoed `/close` — **everything else returns and is discarded.**
- Meanwhile the dedup gate (`src/services/bridge-fallback-gate.ts:286`, `shouldSuppressBridgeEmit`) decides "did I already reply?" purely from markers written by `botmux send` — the three write sites are all in `src/cli.ts` (8833 / 8915 / 9384).

So the gap isn't that we *can't* see a side-channel send; it's that we see it and **don't record it**. Recording a marker from the self-message echo would change the gate's criterion from *"did `botmux send` write a marker?"* to *"did a message from me actually appear in this chat this turn?"* — which is mechanism-agnostic and covers `lark-cli`, `curl`, and any future bypass, **without depending on the model's compliance**.

The marker shape looks accommodating: `BridgeSendMarker` (line 194) has `sentAtMs` + optional `messageId`/`contentLength`, and `markerSetCoversFinal` already treats a marker *without* `contentLength` as conservative-suppress — so even a minimal echo-derived marker would fail safe rather than leak a duplicate.

**One honest caveat, and it's the reason I'm asking rather than asserting:** this only works if the bypass message is attributable to the bot in the echo. On the box I tested, `lark-cli profile list` showed the only `active` profile is a *human user* identity, with the bot apps inactive. That implies two shapes:
- sent via `--as bot --profile <bot appId>` → `senderOpenId == botOpenId`, the echo approach lands cleanly;
- sent via an active *user* profile → the message appears authored by that user (`senderType=user`), and the echo can't be attributed to the bot.

**Do you know which shape produced the duplicate in your report?** The screenshot in #1163 wasn't attached (watermarked), so I couldn't tell. If it's the second, the echo approach needs a wider criterion (any new message under the turn's anchor), which needs care not to count a human's interjection.

## Correcting something in my own earlier reasoning

I had suggested capability isolation (#1163 direction 3, "don't expose `lark-cli` to ordinary chat sessions") as the clean fix. Having looked properly, **that's harder than I made it sound**: `lark-cli` is a real dependency — `src/vc-agent/polling-source.ts` spawns it for VC meetings, and there's a family of `lark-doc` / `lark-base` / `lark-sheets` skills relying on it. Blanket removal would break working features; it would have to be scoped to chat-send subcommands specifically. Still viable, but a much wider blast radius than touching the gate.

## What I'd like from you

1. **Your read on the echo-marker approach** — does it look right to you, and did I miss a reason it can't work? You've been closer to this failure than I have.
2. **Which identity shape** your reproduction used (bot profile vs user profile), if you know.
3. **Whether you'd want to take the gate-side change yourself** — you already own the diagnosis, and I'd rather ask than assume. If you'd prefer not to, that's completely fine; it can be a separate issue for whoever picks it up.
4. One small request regardless of the outcome: consider changing `Fixes #1163` to `Refs #1163`. As written, merging auto-closes the issue while directions 3 and 4 are still open — and direction 4 is the one that actually closes the detection gap.

To be clear about status: no decision has been made, and this isn't a rejection — the maintainer is weighing whether to take this as an interim safety net, take only the hook-injected variant (which is the one config can't override), or go straight for the gate fix. Your input is likely to decide it, which is why I'm asking before anyone acts.

Thanks again for the diagnosis — pinpointing that a bypass leaves no marker and therefore defeats the fallback check is the insight the real fix will be built on.

---

*Automated preliminary review; the maintainer has the final say.*

mvanhorn:
Agreed on the prompt clause, and thanks for the careful second look.

On (1), the echo-marker approach: I think it's right, and it's better than what I proposed. The reason I like it is the one you gave. It moves the gate's question from "did the sanctioned code path run" to "did a message from the bot actually land in this chat", which is what the fallback actually cares about. My clause only re-states a rule the model already had and ignored, so it cannot bind a model that reaches for curl or an MCP tool instead.

Two things I would want to check before calling it done, neither fatal:

Ordering. The echo is asynchronous. If `shouldSuppressBridgeEmit` runs before Lark delivers `im.message.receive_v1` for the side-channel send, the marker is not there yet and we still double-send. That turns a correctness fix into a race we win most of the time. Worth measuring echo latency against the gate's decision point, and if the window is real, either waiting briefly on the anchor or accepting that this closes the common case rather than all of it.

Scope of suppression. "Any bot message under this turn's anchor" is broader than "botmux send already replied". A deliberate second message in one turn would now be suppressed where it is not today. `markerSetCoversFinal` failing conservative helps, but it is worth being explicit that the new criterion changes behavior for legitimate multi-message turns, not only for bypasses.

On (2), identity shape: I cannot answer that confidently from memory, and I would rather not guess on something the design hinges on. I will re-check my reproduction and follow up with the sender identity. If it turns out to be the user-profile shape, I agree the echo cannot attribute it to the bot and the wider anchor criterion is needed, with the human-interjection problem you flagged.

On (3): please treat the gate change as unowned. I am happy for whoever is closest to it to take it, and I do not want to block it by sitting on it. If it is still open once I have confirmed the identity shape, I will pick it up then.

On (4): agreed, I will change `Fixes #1163` to `Refs #1163`. It should not auto-close while directions 3 and 4 are open.

On this PR itself: I would suggest the hook-injected variant only, if the maintainer wants an interim net. It is the one config cannot override, so it is the only part carrying a real guarantee, and it does not spend prompt budget in sessions that do not need it. If the maintainer would rather wait for the gate fix, I have no objection to closing this. The prompt-side argument against it is sound and I do not want to defend a change I now think is weak.

One correction on my side: you are right about the capability-isolation suggestion. `lark-cli` is load-bearing for VC polling and the lark-doc/base/sheets skills, and I underweighted that. Scoping to chat-send subcommands specifically would be the only safe version.

PR #1300, opened 2026-09-07
feat(cli): add MiniMax CLI (mmx) adapter

lijason1121:
Add a `minimax` CLI adapter bridging MiniMax's official CLI (`mmx`, package `mmx-cli`). Its interactive multi-turn surface is `mmx text repl`, a readline-style chat REPL that requires a TTY — provided by the PTY/tmux backend.

The repl is a pure chat/generation loop with no shell or file-tool surface, so like mira/riff it relays via quiescence + screen capture rather than the `botmux send` wrapper: no skillsDir, empty systemHints, altScreen=false (it redraws its input line in place, not via the alternate screen buffer). Auth resolves from `mmx auth login` (~/.mmx/config.json); the whole ~/.mmx dir is the authPath so a sandboxed first login persists.

Registered per src/adapters/cli/CLAUDE.md checklist:
- src/adapters/cli/minimax.ts (new)
- CliId union (types.ts)
- registry.ts: RAW_CLI_EXECUTABLES + import + export + switch case
- worker.ts CLI_DISPLAY_NAMES
- im/lark/card-builder.ts cliDisplayNames
- setup/bot-config-editor.ts CLI_ID_CHOICES (#30, appended) + CLI_DISPLAY_LABELS

Verified: tsc --noEmit clean; cli-adapters + cli-availability unit suites pass (449); end-to-end PTY harness drove a real MiniMax-M3 reply through buildArgs -> spawn -> writeInput -> output.

### DIFF
--- a/README.en.md
+++ b/README.en.md
@@ -81,7 +81,7 @@ More: [Roles & teams](https://deepcoldy.github.io/botmux/en/roles) · [File sand
 
 Switch with `cliId` in `bots.json`. **20+ adapters**, spanning local CLIs (process-isolated, reachable via `tmux attach`) and API / cloud agents (e.g. Mira, riff — reached over API / remote, not a local process; mojo is API-driven but executes tools on the bot host by default, set cloud: true for the remote sandbox). Representative ones:
 
-`claude-code` · `codex` · `gemini` · `cursor` · `opencode` · `opencode2` · `antigravity` · `copilot` · `grok` · `kimi` · `kiro-cli` · `reasonix` · `dsh` · `aiden` · `coco` (TRAE) · `hermes` · `ebsd` · `mira` · `riff` (cloud agent) … · `mojo` (API-driven, host execution by default) …
+`claude-code` · `codex` · `gemini` · `cursor` · `opencode` · `opencode2` · `antigravity` · `copilot` · `grok` · `kimi` · `kiro-cli` · `reasonix` · `dsh` · `aiden` · `coco` (TRAE) · `hermes` · `ebsd` · `mira` · `riff` (cloud agent) … · `mojo` (API-driven, host execution by default) · `minimax` (MiniMax `mmx text repl`; region set by `mmx auth login --region cn|global`, isolate multiple regions on one host via per-bot `env` `MMX_CONFIG_DIR`) …
 
 The `ebsd` adapter uses a dedicated external service identity and native OMP session storage. Operators must provide the Diag Gateway token and ByteCloud service account through permission-restricted files, never through `bots.json`.
 
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ npm 包内带的是**同一个自包含二进制**（按 os/arch 只装匹配的
 
 `bots.json` 里用 `cliId` 一键切换。**20+ 适配器**，覆盖本地 CLI（进程隔离，`tmux attach` 可直连）和 API / 云 Agent（如 Mira、riff——通过 API / 远端接入，非本地进程；mojo 为 API 驱动、默认在宿主机执行工具，可配 cloud: true 走云沙箱）。代表项：
 
-`claude-code` · `codex` · `gemini` · `cursor` · `opencode` · `opencode2` · `antigravity` · `copilot` · `grok` · `kimi` · `kiro-cli` · `reasonix` · `dsh` · `aiden` · `coco`(TRAE) · `hermes` · `ebsd` · `mira` · `riff`(云 Agent) … · `mojo`(API 驱动,默认宿主机执行) …
+`claude-code` · `codex` · `gemini` · `cursor` · `opencode` · `opencode2` · `antigravity` · `copilot` · `grok` · `kimi` · `kiro-cli` · `reasonix` · `dsh` · `aiden` · `coco`(TRAE) · `hermes` · `ebsd` · `mira` · `riff`(云 Agent) … · `mojo`(API 驱动,默认宿主机执行) · `minimax`(MiniMax `mmx text repl`；区域由 `mmx auth login --region cn|global` 决定，同机多区用 per-bot `env` 的 `MMX_CONFIG_DIR` 隔离) …
 
 `ebsd` 使用独立的外部服务身份和原生 OMP 会话目录；部署方必须通过受限权限文件配置 Diag Gateway token 与 ByteCloud service account，不能把密钥写入 `bots.json`。
 
--- a/src/adapters/cli/minimax.ts
+++ b/src/adapters/cli/minimax.ts
@@ -25,6 +25,26 @@ import { delay } from '../../utils/timing.js';
  * dir is the authPath so a sandboxed first login persists). The `--api-key`
  * flag and `MMX_CONFIG_DIR` env are alternative resolution paths; we do not
  * bake a key into argv.
+ *
+ * Region (CN vs. global) — supported, resolved by mmx itself:
+ *   `mmx` picks the API host by precedence `--base-url` > `--region` >
+ *   the `region` field in `~/.mmx/config.json` (written at login, default
+ *   `global`). `cn` → api.minimaxi.com, `global` → api.minimax.io. This
+ *   adapter does NOT pass `--region`, so a bot follows whichever region you
+ *   chose at `mmx auth login`:
+ *     mmx auth login --api-key sk-... --region cn        # China
+ *     mmx auth login --api-key sk-... --region global    # international
+ *
+ *   Running a CN bot and a global bot on the SAME host: `~/.mmx` holds one
+ *   region, so give each bot its own credential dir via the per-bot `env`
+ *   field in bots.json (`MMX_CONFIG_DIR`), each logged into its own region:
+ *     bot-cn:      env: { "MMX_CONFIG_DIR": "~/.mmx-cn" }
+ *     bot-global:  env: { "MMX_CONFIG_DIR": "~/.mmx-global" }
+ *   Prepare each once, e.g.
+ *     MMX_CONFIG_DIR=~/.mmx-cn mmx auth login --api-key sk-... --region cn
+ *   (Under the file sandbox, `authPaths` is the static `~/.mmx`; a
+ *   redirected MMX_CONFIG_DIR would additionally need that dir exposed —
+ *   irrelevant to the default non-sandboxed setup.)
  */
 export function createMinimaxAdapter(pathOverride?: string): CliAdapter {
   // resolvedBin is lazy: setup constructs adapters only to read static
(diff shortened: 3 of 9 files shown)


deepcoldy:
你好！PR #1300 的自动评审群已创建：[加入评审群](https://applink.feishu.cn/client/chat/open?openChatId=oc_35521559ff06a1fa6f8c8f9b0926b36d)

目前你还不在自动拉群名单里，暂时无法自动邀请你入群。如果希望后续 PR 能自动拉你进评审群，请把你的 GitHub 账号和飞书信息补进[这个名单文档](https://bytedance.larkoffice.com/wiki/WJ1nwWbtxi89erkNGNbcgkt9nUe)，补好后后续复审会自动拉你进群。

这是自动流程，感谢贡献！

deepcoldy:
Thanks for a genuinely careful adapter — the `authPaths` directory-level choice, the appended-at-the-tail `CLI_ID_CHOICES` numbering, the lazy `resolvedBin`, and the region/`MMX_CONFIG_DIR` documentation all match the house rules, and the header comment is more thorough than most adapters get. Two blocking issues came out of review, one of which the checklist genuinely did not warn you about.

**1. A registration point outside the checklist makes the unit suite red**

`test/cli-id-roster-derivation.test.ts` asserts two-way key equality between `ALL_CLI_IDS` and its local `EXPECTED` map (it deliberately does this at runtime, because `test/` is not covered by `tsconfig`, so the `Record<CliId, …>` annotation is not enforced). `minimax` is missing there:

```
AssertionError: stale or missing entries in EXPECTED: -"minimax"
```

The fix is one line next to the other no-skill CLIs:

```ts
// Tool-less chat REPL (`mmx text repl`) — no skills dir of its own.
minimax: 'none',
```

Baseline check: the same file is 6/6 green on clean `master` and 1-failed with this PR, so it is caused by the change rather than flaky. Your reported `449 passed` is accurate — `cli-adapters` + `cli-availability` really do pass; that file just isn't in those two. Worth noting this step is **not** in `src/adapters/cli/CLAUDE.md`'s 7-step checklist, so following the docs would still miss it — it's probably worth adding as step 8 so the next adapter doesn't repeat it. Also, CI never actually ran here (fork runs are parked at `action_required`), so nothing flagged it automatically.

**2. `systemHints: []` is not the switch that suppresses the routing block**

The comment says hints are omitted "same as mira / riff / mojo", but those three additionally set `injectsSessionContext: true`, and *that* is the flag `session-manager.ts:1145` gates on:

```ts
const hints = adapter.injectsSessionContext ? [] : buildBotmuxShellHints(...)
```

`adapter.systemHints` has no reader in `src/` at all (only tests, plus a comment noting it was superseded by the i18n path). Measured with `buildNewTopicPrompt`:

```
minimax  routing=true  botmux_send=true  session_id=true  len=1681
mira     routing=false botmux_send=false session_id=false len=33
```

So a tool-less chat REPL currently receives the full `<botmux_routing>` block instructing it to reply via `botmux send` and to pick one of `--mention`/`--mention-back`/`--no-mention` — none of which it can do.

**2b. Compounding this: the first prompt is multi-line, and delivery splits it per line**

`writeInput` is copied from hermes, but hermes drives a TUI composer while `mmx text repl` is a readline-style prompt. `tmux send-keys -l` sends `\n` as Enter (`tmux-backend.ts` says so in its own comment), and `PtyBackend` has no `sendText`, so the `else` branch (`write(content)` + `'\r'`) applies there. Measured on both transports with a Node readline consumer on an isolated tmux socket: a 3-line message arrives as three independent submissions. The assembled first prompt is 24 lines, so submission #1 is just `<botmux_routing>` and the user's actual question lands around line 23.

One caveat on the fix, which we checked before suggesting it: **switching to `pasteText` is not sufficient by itself.** `tmux paste-buffer -p` only inserts `\x1b[200~ … \x1b[201~` when the pane has requested bracketed-paste mode. Node's built-in `readline` never emits `\x1b[?2004h`, and with `load-buffer` + `paste-buffer -d -p` the same 3-line input still arrived as 3 submissions; with a consumer that does request it, the identical command delivered one bracketed chunk. So whether `pasteText` helps depends on how `mmx`'s input layer is implemented — if it is plain `readline`, a different approach is needed (e.g. collapsing to a single line, or sending in parts and waiting on echo). Since you have a working `mmx` install and we don't, could you verify a genuinely multi-line first prompt end-to-end on the real CLI? That is the part we cannot check from here.

**Suggested scope to unblock:**
1. Add `minimax: 'none'` to the roster test's `EXPECTED`.
2. Verify multi-line first-turn delivery against real `mmx`, and fix the transport accordingly.
3. Decide the `injectsSessionContext` stance explicitly and correct the comment either way — if minimax should behave like mira/riff, set the flag; if it should keep routing hints, the comment needs updating and the delivery fix in (2) becomes the load-bearing part.

For transparency: everything above was verified on a local rebase onto latest `master` (`506b52d05`), which drifted twice during review; your patch applies cleanly with no conflicts and no overlap with the drift. This is automated preliminary review feedback — a maintainer's own review is the final word, and points 2/3 in particular are worth their judgement rather than ours.

deepcoldy:
Thanks for the quick turnaround on the roster entry — that one is properly fixed, not just silenced. We verified the classification is factually right (`skillsDir` is `undefined` and `resolveSkillInjectionSupport('minimax')` genuinely returns `'none'`), and that the new assertion is load-bearing: flipping it to `'global'` turns the test red. The commit is a clean pure insertion, `6/6` on that file, `455/455` across `cli-adapters` + `cli-availability` + the roster suite, `tsc` clean. Verified on a local rebase onto current `master` (`0ad77ffcc`), which applies with no conflicts.

Full-suite check: 35 failures on the PR vs 33 on the same-base `master`. The two PR-only ones (`hook-runner.test.ts`, `tmux-pipe-backend-exit.test.ts`) are in files this PR touches neither the tests nor the sources of, and both pass 3/3 in isolation — concurrency flakes, not regressions from your change.

**The remaining item is a merge blocker, not a maintainer-pending question** — we want to be explicit about that, since it's easy to read the earlier review as "fix the red one, leave the rest for maintainers."

`src/adapters/cli/minimax.ts` is byte-identical to the previous head, so the multi-line first-turn delivery issue stands exactly as measured:

- `injectsSessionContext` is still unset, so the adapter still receives the full `<botmux_routing>` block (measured `routing=true`, `botmux_send=true`) — instructions to shell out to `botmux send`, which this CLI cannot do.
- The assembled first prompt is **24 lines** (routing block + skill catalog + the user's message). With `sendText` (`tmux send-keys -l`), `\n` is Enter, so a readline-style prompt receives it as one submission per line: submission #1 is just `<botmux_routing>` and the user's actual question lands around line 23.

We suspect the e2e harness in your PR description used a single-line prompt, which wouldn't surface this — that's why we're asking specifically for the multi-line case.

**What would unblock it:** run a genuinely multi-line first turn against your real `mmx` install and share the evidence (transcript, log, or recording — whatever is easiest). Any of these outcomes is fine:

- If it arrives as **one** submission, we're wrong about the impact and will drop this to non-blocking immediately — we can't test it ourselves because we have no `mmx` here.
- If it splits per line, the fix depends on `mmx`'s input layer: if it handles bracketed paste (a custom raw-mode parser, like Ink-based CLIs), routing multi-line content through `pasteText` is enough. If it's plain Node `readline`, `pasteText` will **not** help — we tested this with a readline consumer, and `paste-buffer -p` only inserts the `\x1b[200~`/`\x1b[201~` markers when the application has requested bracketed-paste mode, which `readline` never does. In that case a different approach is needed and we'd welcome your proposal, since you know the CLI best.

The third item is small and doesn't need to wait on any of the above: decide the `injectsSessionContext` stance and make the comment match. Right now line 92 says hints are omitted "same as mira / riff / mojo", but those three all set `injectsSessionContext: true` and minimax doesn't — so the comment currently states the opposite of the behavior, whichever way you decide to go.

As before: automated preliminary review feedback, and a maintainer's own review is the final word.

deepcoldy:
Both blocking items are resolved — verified by measurement here, not taken from the commit message. Thank you for going and testing the real CLI; that turned out to matter.

**Your real-machine finding corrected ours.** We had measured "3 lines → 3 submissions" against a stock Node `readline` stand-in, because we have no `mmx` on this box. Your mmx 1.0.25 result — interior `\n`/`\r` swallowed, one submission with the lines jammed together — is the accurate one, and we reproduced it locally by modelling that behavior. Both readings are internally correct; they just describe different readline configurations, and yours is the actual CLI. Your version of the bug is also the nastier one: silent content corruption rather than a visible split.

What we verified on a local rebase onto current `master` (`8880fcecb`, applies cleanly):

- **`injectsSessionContext: true`** — `buildNewTopicPrompt('minimax')` now yields only the `<user_message>` envelope, byte-identical in shape to mira/riff/mojo. No `<botmux_routing>`, no `<identity>`, no `<session_id>`, no `botmux send`. Previously it was a 1681-char routing block.
- **Newline folding** — the real assembled first prompt goes 24 lines → 1, with clean content: `"<user_message> 帮我写一首诗 </user_message>"`.
- **Robust across both input models** — we replayed the payload through both a "swallows `\n`" consumer and a stock `readline` consumer. Before the fix: 1 jammed submission and 3 split submissions respectively. After the fix: **one clean, correctly spaced submission in both**. So the fix doesn't depend on which model is right about mmx.
- `tsc` clean; `cli-adapters` + `cli-availability` + roster **459/459** (same-base `master` is 457; the +2 are minimax's own generic cases). Full-suite check shows no regressions attributable to this PR.

Two **non-blocking** follow-ups — neither should hold up the merge, and the maintainer may reasonably skip both:

1. **The two fixes have no test coverage.** Reverting either one leaves the suite fully green: dropping `injectsSessionContext` regresses the behavior (routing block returns) at 459/459 passing, and replacing the fold with `const flattened = content` likewise stays green. Both lines are reachable and load-bearing, so this is a coverage gap rather than dead code — worth a named minimax case so a future edit can't silently undo either. There's an existing pattern to copy at `test/cli-adapters.test.ts:1266` and `test/seed-adapter.test.ts:95`.

2. **The fold regex misses a bare `\r`.** `/\s*\n\s*/` only matches runs containing `\n`, so `"a\rb"` passes through unchanged. Whether that bites depends on how mmx treats a lone `\r` — you're better placed to say. Either way `/\s*[\r\n]+\s*/` is strictly better under both models (a separator when `\r` is swallowed; no double submit when it isn't), and it's a cheap hardening. Note this gap is shared with hermes/reasonix/dsh-tui/kimi, so it's a pre-existing pattern rather than something this PR introduced.

As before: automated preliminary review, and the maintainer's own review is the final word. Our recommendation is that this is technically ready to merge, with the two items above as optional follow-ups.

deepcoldy (approved):
双审通过：三项阻断均已修复并逐项实测验证（injectsSessionContext 抑制生效、多行首投折叠对两种 readline 模型均鲁棒、roster 分类属实且断言承重）。无 breaking change：现有 30 个 CliId 成员逐字保留、顺序不变，各清单零丢失，现存 CLI 行为未触碰。required checks（build / test）全绿；bun-test 为 continue-on-error 的非必需腿，其唯一红点 child-env.test.ts 是干净 master 上同样复现的既有 node-pty lost-read flake，与本 PR 无关（本 PR 零触碰该文件）。

PR #1347, opened 2026-09-09
fix(mojo): isolate PM2 metadata from child environments

swtxbling:
## Summary

- scrub PM2-managed process metadata from the ambient environment before spawning Mojo children
- preserve PATH, HOME, proxies, certificates, credentials, and Botmux session routing; explicit bot env and mojo.env keep their existing precedence
- use the same effective environment for worker wrapper resolution and the real MojoBackend spawn path
- cover host, cloud, sandbox, dynamic instance_var safety, explicit overrides, and non-Mojo compatibility
- update the bug ledger with real-device verification and final review evidence

## Validation

- Mojo 1.0.12 direct host Bash: 2/2 completed, one tool call each, return code 0
- real MojoBackend host smoke: completed with Bash result and final answer
- full Mojo Vitest suite: 33 files passed, 1 skipped; 667 tests passed, 32 skipped
- focused independent review: 140 passed, 1 skipped across the directly relevant suites
- bun 1.4.2 build: passed, including TypeScript checks, dashboard bundle, dist audit, and embedded-asset audit
- git diff --check: passed
- independent final review: PASS, blocker 0, major 0, minor 0 after ledger closeout

## Compatibility and evidence boundary

- non-Mojo environment merging is unchanged
- ambient proxy variables remain available and can still be overridden or cleared explicitly
- PM2 Common.js drift coverage runs when PM2 is installed; static fixtures always cover the scrub behavior now that PM2 is no longer a runtime dependency
- on macOS, the isolated smoke workspace has no remaining process using it, but one synthetic-session containment handle remains fail-closed because Darwin cannot provide Linux /proc-level process-tree proof

## Tracking

- BUG-20260828-001


swtxbling:
关闭本 PR。复验使用真实 Mojo 1.0.12，在保留代表性 PM2 元数据且完全不启用本分支清洗逻辑的情况下，host Bash `pwd` 连续两轮均正常完成（return code 0、result completed）；干净环境对照同样通过。现有证据表明用户可见故障已由 Mojo 1.0.12 修复，这套 Botmux/Mojo 专属 PM2 静态清洗没有可证明的额外收益，因此不合入。主线中的 legacy PM2 reaper 仍继续保留，用于升级迁移和防止旧、新 fleet 双跑。感谢这轮定位、实现和多轮验证。

PR #1417, opened 2026-09-16
fix(lark): 串行化实时卡片显示模式更新

kunshanyep-beep:
## 问题

当实时卡片已有 PATCH 在途时，显示模式切换会同时进入 PATCH 队列并通过同步 callback 返回完整卡片。飞书可能先应用 callback，再让较旧的 PATCH 覆盖回来，造成显示状态回退。

修复后还需兼容队列拒绝更新的路径：聊天范围的替身轮次、运行时关闭实时卡片、无飞书传输或卡片 ID 尚不可用时，如果只返回 toast，点击后的卡片会永久保持旧状态。

## 改动

- `scheduleCardPatch` 返回是否接受本次发布：立即发送或成功入队返回 `true`，所有保护分支返回 `false`。
- 当前实时卡片的切换在队列接受时只返回 toast，避免队列外的第二次同步更新。
- 队列拒绝时回退为同步返回重建卡片，确保点击的卡片仍反映新的显示模式。
- 保留点击历史卡片时原有的直接更新与同步返回行为。
- 增加替身轮次拒绝 PATCH 队列的回归测试，并锁定无传输场景的布尔返回契约。

## 影响范围

改动仅涉及飞书实时会话卡片的显示模式切换与卡片 PATCH 调度；不改变其它 IM、CLI、后端或会话生命周期。替身轮次和运行时卡片配置关闭场景现在会使用 callback 更新被点击卡片；正常实时卡片继续走单一串行队列。

本次不改变卡片布局或视觉样式，仅调整更新时序和回调载荷，因此没有可展示的截图差异。

## 验证

- `vitest run --project unit test/card-integration.test.ts`（41 passed）
- 相关 unit：`card-integration`、`api-only-card-patch-suppression`、`api-only-mode-wiring`、`substitute-control-card`（109 passed）
- `vitest run --project e2e test/card-toggle.e2e.ts`（8 passed）
- `bun run build`


deepcoldy:
Hi — thanks for this fix. The race analysis is correct: when a toggle click arrives while an older card PATCH is still in flight, returning the rebuilt card from the synchronous callback creates a second, out-of-queue update, and Lark can apply the callback result first and let the older PATCH restore stale display state afterwards. Routing the live card exclusively through `scheduleCardPatch` and ACKing with a toast-only response addresses that. I also verified the two changed test assertions genuinely fail when the fix is reverted (mutation check), so the regression coverage is real.

One blocking issue from automated review (preliminary — final call is the maintainer's):

**`scheduleCardPatch` can silently decline to publish, and in that case the toast-only reply leaves the clicked card permanently unchanged.**

`scheduleCardPatch` has several early-return no-op paths (`src/core/worker-pool.ts`):

1. `streamingCardDisabled(ds, turnId)` — bot/chat with live card disabled, or a substitute turn (`isSubstituteTurn`);
2. `larkTransportEnabled(...)` false;
3. `!ds.streamCardId || ds.streamCardId === CARD_POSTING_SENTINEL` — the window during which a new-turn card is being POSTed (`__posting__`). This one happens on every new turn: a toggle click during that POST now gets only a toast, with zero API PATCHes and nothing queued (`pendingCardJson` stays null).

Before this PR, these paths were rescued by the synchronous card return — Lark replaced the clicked card with the rebuilt body regardless of the queue. After this PR the callback returns only a toast, the queue never owns a publish, and the visible card keeps its stale display mode (even though `ds.displayMode` was already flipped in memory).

I reproduced it with a probe on top of latest `master` (rebased, no conflicts): for both the `disableStreamingCard` and `__posting__` cases the PR returns `{toast:...}` with `patches=0, pending=null`, while master returns the rebuilt `{type:"streaming",...}` card synchronously.

Suggested direction (I prototyped it locally — tsc clean, the 40 card-integration tests and the 8 toggle e2e tests all pass, and the probe flips back to a synchronous card on all three decline paths): make `scheduleCardPatch` return a boolean indicating whether it actually accepted the publish (`true` for immediate flush or queued, `false` for every early-return guard), and in the toggle handler fall back to the synchronous card return when it declines:

```ts
} else if (!scheduleCardPatch(ds, cardJson)) {
  // Queue declined (card disabled / mid-POST / no transport): publish via the
  // synchronous callback so the clicked card still reflects the toggle.
  logger.info(`[${tag(ds)}] Display mode → ${next}`);
  try { return JSON.parse(cardJson); } catch { /* fall through */ }
  return;
}
```

Two non-blocking notes:
- The same `scheduleCardPatch(...)` + synchronous `return JSON.parse(cardJson)` pattern still exists in `retry_last_task`, `refresh_screenshot`, and `term_action`. The race this PR fixes theoretically applies there too (though `refresh_screenshot` deliberately returns synchronously to avoid an image flash while the fresh screenshot PATCH is in flight). Might be worth a follow-up to decide on one consistent strategy.
- Consider adding a regression assertion for the decline path (queue refuses → callback still carries the rebuilt card), so this fallback stays locked.

This is an automated preliminary review; the maintainer's review takes precedence. Happy to help verify a revised push.

deepcoldy:
Follow-up correction to my earlier comment after a secondary review pass (same automated review process):

**Reachability refinement — the `__posting__` case I cited as "happens on every new turn" is not reachable from a genuine click.** Real card-action callbacks carry `context.open_message_id` (the clicked card physically exists, so its id can never be the in-flight `__posting__` sentinel). When a new-turn card is mid-POST and the user clicks the previously posted card, `cardMessageId !== ds.streamCardId` routes into the existing legacy-card migration branch (direct PATCH + synchronous card return), not the `else`/`scheduleCardPatch` branch. The toast-only drop on that branch only occurs when `cardMessageId` is absent — which today only happens in synthetically constructed test events.

The blocking finding still stands, but the real-world exposure is narrower:

1. **Substitute ("avatar") turns in chat scope (primary case).** A toggle click during an active substitute turn calls `scheduleCardPatch(ds, cardJson)` with no turnId, `isSubstituteTurn` returns true, the queue declines → toast only, the visible card never updates. (Verified by probe A/B against `master`: master synchronously returns the rebuilt card; the PR returns the toast with zero PATCHes.)
2. **Live-config change (secondary).** `disableStreamingCard` / `noCardChats` flipped on while a previously posted streaming card is still clickable.

One more implementation note for the suggested fix: changing the guard to `return false` also needs a one-line marker update in `test/api-only-mode-wiring.test.ts` (the `region(...)` assertion ends on the literal text `if (streamingCardDisabled(ds, turnId)) return;`) — verified that it fails on the old marker and passes with the marker updated.

Everything above was re-verified on the latest `master` (rebased, clean). Still a preliminary automated review; the maintainer's review takes precedence.

kunshanyep-beep:
已处理阻断项，更新在 `dd7ffeb7`：

- `scheduleCardPatch` 现在返回是否接受发布；传输关闭、实时卡片禁用、卡片 ID 缺失或处于 `__posting__` 时返回 `false`，立即发送/入队返回 `true`。
- toggle handler 仅在队列接受时返回 toast；队列拒绝时同步返回重建卡片，保留替身轮次和运行时配置切换下的可见更新。
- 新增 chat-scope 替身轮次回归用例，并更新 API-only/source-lock 测试。

本地验证：card integration 41 passed、相关 unit 109 passed、toggle e2e 8 passed、`bun run build` 通过；本次 GitHub CI 全部通过。

deepcoldy (approved):
双审（Claude + pi）0 阻断收敛；rebase 最新 master 验证 tsc/build/463 测试全绿，变异有效。

PR #1418, opened 2026-09-16
fix(feedback): handle shared and stale card callbacks

kunshanyep-beep:
## Problem

Without a platform-provided interaction ID, feedback deduplication could only compare the submitted value with the latest stored value. In an `upvote -> negative -> delayed old upvote` sequence, the delayed callback was no longer consecutive and could create a third revision that restored the old choice.

The first implementation also treated a missing card version as revision zero. That made shared cards accept only the first reviewer's click, and legacy negative-feedback rows with a `NULL` semantic did not take the deferred update path.

## Change

- Put the feedback revision ID into every rendered feedback action as an optimistic card version.
- Check that version inside the same SQLite write transaction before appending a revision.
- Treat an absent version as no precondition, so every reviewer can use the original shared card.
- Make callback keys depend on the card version instead of mutable server state.
- Return a visible toast for stale or unknown versions and repaint the latest delivery state.
- Preserve that toast while negative feedback is acknowledged immediately and patched asynchronously.
- Resolve the rendered semantic from the delivery's frozen button policy, so legacy rows with a `NULL` semantic still use the deferred update path.
- Keep non-card `recordFeedback` callers backward-compatible by making the precondition optional.

## Validation

- 14 feedback and event-dispatcher test files: 481 tests passed
- Added regressions for shared-card reviewers, stale and unknown versions, legacy `NULL` semantic rows, and deferred ACK toasts
- Re-ran all five files that initially failed locally with Bun 1.4.2 on `PATH`: 150 tests passed; the sole remaining `/tw` assertion also fails unchanged on `origin/master`
- `bun run build` with Bun 1.4.2


deepcoldy:
Thanks for this — the direction is right, and the bug you describe in the PR body is real. I verified it: on current trunk the `upvote → negative → delayed old upvote` sequence writes a **third revision and restores the stale upvote**, while your branch stops at 2 revisions and keeps `incorrect`. That core fix works.

Two things look worth another pass before merge, plus one smaller hygiene item.

### 1. The first card a user sees carries no version token, and its absence is read as an assertion

All three send sites call `buildFeedbackElement(policy)` with no state (`src/cli.ts:10553`, `src/im/lark/md-card.ts:1267`, `:1354`), so `versionedValue` returns the value unchanged and the originally-sent card has no `expected_feedback_id`. The handler then maps that absence to `?? null`, and in the store `null` is a meaningful assertion — "this delivery currently has zero revisions."

The consequence shows up on a shared card. With the default policy and `audience: 'reviewers'` (three reviewers, the multi-reviewer case the policy docs describe), each reviewer clicking the card they were actually sent gives:

```
ou_r1 -> recorded=1  toast=none
ou_r2 -> recorded=0  toast=none
ou_r3 -> recorded=0  toast=none
```

Two of three votes are dropped **with no toast at all**, and each clicker is shown someone else's choice. Rejecting those later clicks is defensible for optimistic concurrency, but silently is not.

The root cause of the silence is that `recordFeedback` gained `status: 'stale'` and the handler never reads it — `recorded.status` has no consumer anywhere in the repo, so a rejected write and a successful one are indistinguishable at the call site and both repaint the card. Reading `status` and returning a toast (plus the freshly repainted card) would cover this. Optionally, passing `undefined` rather than `?? null` when the card carried no token would let that first click simply succeed — if you do that, the `?? null` inside `callbackKey` needs the same treatment so the hash stays consistent.

### 2. Deferred-card detection now depends on a column that was never backfilled

The condition changed to `recorded.feedback.semantic === 'negative'`, which reads the stored `semantic` column. `MIGRATE_V4_TO_V5` adds that column with a bare `ALTER TABLE ... ADD COLUMN semantic TEXT` — no default, no backfill — so rows written before the migration have `semantic = NULL`. `mapFeedback` maps that to `undefined`, the comparison is false, and a negative selection returns `{card}` instead of `{deferredCard}`, losing the ACK-then-patch that keeps Lark from restoring the pre-click card.

Reproduced with a pre-v5 row (`semantic` NULL): `deferredCard=false card=true`, where trunk gives `deferredCard=true`.

One caveat if you fix this by going back to a policy lookup: keying it off the **clicked** `result` breaks your own `does not let a redelivered old choice overwrite a newer selection` test, because on a stale replay the card being rendered is the *stored* state, not the clicked one. Keying it off the rendered row's result works — I tried this locally:

```ts
const renderedSemantic = delivery.policy.buttons.find(item => item.key === recorded.feedback.result)?.semantic;
if (action === 'feedback_submit' && renderedSemantic === 'negative') {
```

That restores deferral for legacy NULL rows and keeps all 34 tests in the three feedback suites green (17 feedback-related files, 253/253; `tsc --noEmit` clean).

### 3. `feedback_version_mismatch` is thrown rather than returned

When a card carries a token but the store has no matching revision, `if (!latest) throw new Error('feedback_version_mismatch')` propagates out of `handleSkillFeedbackCardAction`. The sole call site (`src/im/lark/card-handler.ts:1454`) doesn't handle it, so the click disappears with only a server-side log. It's an edge case and low severity, but returning a toast like the other validation branches would be more consistent.

### One note on what *not* to change

The delivery-wide `latest` query (no `operator_subject_id`) initially looked wrong to me, and I want to flag explicitly that it appears **correct** for a shared card, so it doesn't get "fixed" by mistake. I tried making it per-operator: because a repainted card carries the last clicker's token, the next person's click then compares that token against their own (empty) history and gets rejected — in my run it threw `feedback_version_mismatch` for a click that currently succeeds. The same reasoning applies to the duplicate branch returning the delivery-level latest: that's what keeps a delayed replay from repainting the card back to the old choice, which is exactly the bug this PR fixes.

---

Verification notes: rebased locally onto current trunk (`dc7b4e636`) — clean, no conflicts, and the PR's own net diff is byte-identical before and after, so purely mechanical. Trunk has since touched `skill-feedback-card.ts`, but only in the requester identity gate, upstream of these changes. Reverting just the two source files to trunk while keeping your new tests turns exactly 3 tests red, so the tests do constrain the fix. The wider unit-suite failures are identical on both trees (same 9 files) and unrelated to this PR.

This is an automated preliminary review, so please treat it as input rather than a verdict — some of it may be wrong, and the maintainer's review is what counts.

kunshanyep-beep:
Addressed all three review points in `320d3ce99`:

1. A missing card version now stays `undefined` (including in the callback key), so it means “no precondition” rather than “zero revisions”. Multiple reviewers can therefore click the original shared card independently. A genuinely stale version returns a warning toast and repaints the latest state.
2. Deferred-card selection now derives the rendered semantic from `delivery.policy.buttons` using the recorded result, so legacy rows whose stored `semantic` is `NULL` still follow the negative-feedback deferred path.
3. `feedback_version_mismatch` is caught at the card-handler boundary and returned as a visible error toast instead of disappearing into the server log. The dispatcher also preserves a toast while scheduling a deferred card patch.

The delivery-wide latest query is intentionally unchanged because shared cards need that behavior.

Validation: 481 focused feedback/dispatcher tests pass, the Bun 1.4.2 build passes, and the new regressions cover shared reviewers, legacy `NULL` semantic rows, stale/unknown versions, and deferred ACK toasts.

kunshanyep-beep:
CI update: all build, binary, Bun, and the other two Node test shards passed. `test (1/3)` has one unrelated timing failure in `test/statusline-cli.test.ts`: the 12-second watchdog returned in 12088ms, 88ms over its assertion threshold; the other 8622 tests in that shard passed. The complete statusline file passes locally (7/7). I attempted to rerun only the failed jobs, but GitHub requires repository admin rights for that action on this fork PR.

deepcoldy (approved):
双审 0 阻断：共享卡丢票/陈旧回调/老数据 deferred 均已修复并经探针、变异对照与合并结果树（tsc 0、反馈 18 文件 + event-dispatcher 669 全绿）验证。

PR #1421, opened 2026-09-16
fix(daemon): break bot↔bot cross-principal auto-reply storms

Lotu527:
## Problem

Two bots that both run with `mentionMode: always` can lock a thread into an infinite auto-reply loop.

When bot A is mid-task and bot B sends a message, the cross-principal interruption flow stages the message and notifies B to classify it ("独立任务" / "对当前任务的建议") — a notice that **@-mentions B**. B (`mentionMode: always`) auto-replies to that notice; the reply is itself a fresh cross-principal message to A, which notifies again, and so on.

Observed live: two bots sprayed the same "消息已安全暂存…请选择" / "未选择处理方式…" pair dozens of times in seconds until an operator restarted botmux. Ground truth from the session stores confirmed the mutual loop (`quoteTargetSenderIsBot=true`, `turnReplyContexts` of 36 and 48 turns, `lastUserPrompt` = "[来自 <other bot> 的 @mention]…").

#1406 gives a bot proposer a classify deadline, which bounds a *single* stall, but it still @-mentions the bot every round, so two `always` bots keep re-triggering each other.

## Fix

A per-session circuit breaker keyed on the proposer's already-tracked `senderType`:

- `noteCrossPrincipalProposer` counts **consecutive bot** proposers and resets to zero on any human proposer. Above `CROSS_PRINCIPAL_BOT_LOOP_THRESHOLD` (1) it reports `suppressAckPrompt`.
- When it trips, the staged record is marked `loopSuppressed`: it is still persisted durably (nothing is silently lost), but the bot-classify notice is skipped and the drive loop resolves the record silently in the `awaiting_classification` phase — no notice, no @-mention — so neither bot is re-poked and the storm dies.
- A human proposer resets the counter, so ordinary human interruptions are never suppressed, and a person can always resend to restore the normal notice flow.

Suppression fires when the counter is strictly greater than the threshold, so at `1` the **first** bot interruption in a consecutive run still gets exactly one classify notice and every later one is silenced until a human breaks the run — one prompt, no repeat @-mentions. (Originally `3`; lowered to `1` after review — cards 2–3 were pure noise, since a cooperating bot classifies via `--as` without advancing the counter and only a non-classifying looping bot ever reached counts 2–3.)

## Tests

Pure logic lives in `cross-principal-interruption-store.ts` and is unit-tested:
- a human proposer is never suppressed and keeps the counter at zero;
- suppression trips only *above* the threshold and stays tripped while the bot keeps interrupting;
- a human proposer mid-storm resets the counter so the next sequence gets full runway again;
- the threshold value is pinned to `1` (prompt on the 1st bot interruption, suppress from the 2nd) with a literal assertion, so a future edit of the constant fails the suite instead of self-adapting.

`tsc --noEmit` clean; `vitest` cross-principal + v3-loop suites green (50 tests, including #1406's `cross-principal-choice` tests).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


deepcoldy:
你好 @Lotu527 ，这个 PR 的自动评审群已建好：[飞书评审群](https://applink.feishu.cn/client/chat/open?openChatId=oc_d8897576bef3f5e4dae7750304f983fe)，首次 review 与复审正在群内进行。不过你目前不在我们的自动拉群名单里（名单中有账号但未绑定飞书信息），暂时没有把你拉进群。如果希望直接在群里参与讨论，麻烦在这份名单文档里把你的 GitHub 账号和飞书信息补全：https://bytedance.larkoffice.com/wiki/WJ1nwWbtxi89erkNGNbcgkt9nUe ，补好后后续复审我们会自动把你拉进群。本条为自动流程消息，感谢贡献！

deepcoldy:
Thanks for the PR — the circuit-breaker idea is sound and the core counter logic is correct. This is an **automated preliminary review** against the latest `origin/master` (rebased locally, no conflicts); **the maintainer's review is final**. We found **no blocking correctness issues**, just a few non-blocking refinements.

**Verified**
- The count only advances on genuinely-new, unclassified bot proposers: duplicate redeliveries (`inserted=false`), `--as`-classified messages, and answers to an existing card all return before the counter moves, so retries can't inflate it.
- A suppressed record is removed on the first line of the `awaiting_classification` drive phase, so no terminal @-mention re-pokes the looping bot; and a bot that explicitly sends `botmux send --as ...` still settles and runs that record (`findPendingCrossPrincipalForProposer` does not filter `loopSuppressed`).
- Only the post-#1406 safe field `botClassifyDeadlineAt` is written; the legacy `classificationDeadlineAt`/`ownerDeadlineAt` quarantine probes are untouched, so there is no #1406-style session quarantine.
- Only `senderType === 'bot'` counts; an absent/unknown sender takes the non-bot reset branch (fails open).
- Since #1424 the entire XPI path is behind the default-off experimental switch; this PR adds no new gate and sits fully downstream of it, so it stays dormant unless XPI is enabled — it is good defense-in-depth for when XPI is turned back on.

**Suggested changes (all non-blocking)**
1. **Main one — the counter is a never-decaying, persisted latch.** It resets only on a human cross-principal message — not when the interruption queue drains, the turn ends, a bot successfully classifies, or with passing time. A long-lived session that accumulates 4 unclassified bot interruptions over its lifetime (possibly across days/restarts) would then silently drop all later bot cross-principal messages until a human happens to chime in. We'd suggest resetting the counter when the XPI queue drains (e.g. in `removeCrossPrincipalRecord` when it reaches length 0). This does not weaken the breaker: during a live storm each @-notice elicits another record, so the queue cannot drain; once suppressed there are no new @-notices and the remaining records are finite timeout-terminals whose replies are also suppressed (we checked slow storms and mid-storm restarts). Resetting on successful classification, or adding a time window, would also work.
2. **The two daemon wiring points have no behavioral test.** Mutation check: forcing the suppression gate to `false`, and forcing the drive-loop silent removal to `false`, both leave the suite green (the pure counter function is covered). Could add tests via the existing `__testOnly_driveCrossPrincipalInterruptions` harness (see `daemon-ordinary-ingress-failure-notice.test.ts` for the mocks): (a) the 4th consecutive bot proposer sends no classify notice and persists `loopSuppressed`; (b) the drive loop removes it with no terminal notice; (c) a human proposer resets the counter so the next bot proposer gets the normal notice; (d) after suppression, an `--as` from the proposer bot still settles that record.
3. **Wording nit.** Comments say a "human proposer" resets the counter, but the else branch runs for every non-bot proposer, including `senderType` absent/unknown. Something like "non-bot (human or unknown)" matches the `TrustedCaller` contract (absent is unknown, never human).
4. **Optional — pin the threshold value.** The test loop is bound to the imported `CROSS_PRINCIPAL_BOT_LOOP_THRESHOLD`, so changing the constant alone self-adapts (the strict-vs-`>=` off-by-one direction is covered). Use a literal `3` in the test if you want the value locked.

Happy to see this land after the tweaks; final call is the maintainer's.

Lotu527:
**Follow-up (automated, per maintainer review): lowered the loop-breaker threshold 3 → 1.**

Review conclusion (2026-09-18): the classification prompt should go out at most once per consecutive-bot-interruption run — at threshold 3 the breaker let three "请选择独立任务/建议" cards out before tripping, and cards 2–3 are pure noise (a cooperating bot classifies via `--as`, which never advances the counter, so only a non-classifying looping bot ever reached counts 2–3).

Commit `ee44d51` on this branch:
- `CROSS_PRINCIPAL_BOT_LOOP_THRESHOLD` 3 → 1. Suppression fires when the counter is strictly greater than the threshold, so the **first** bot interruption still gets exactly one classify notice and every later one is silenced until a human breaks the run. No mechanism change; the durable staging and the `--as` escape hatch are untouched.
- Added a literal-pinned test (prompt-on-1st / suppress-from-2nd). The existing breaker tests loop over the imported constant and self-adapt to any value, so the pinned case is what actually locks the value — mutation-checked: flipping the constant back to 3 reds only the new test.

Verified: `tsc --noEmit` clean; the pure-logic store suite is 8/8; daemon-side XPI suites (cross-principal-choice, ordinary-ingress-failure-notice, xpi-shared-cwd-admission / -daemon-callbacks, xpi-session-store-busy-retry) are 54/54.

Two honest notes on what threshold=1 sharpens (both are inherent to "send once", not defects in this commit; and the whole XPI path is behind the default-off switch since #1424):
- The consecutive-bot counter is **session-global, not per-bot-pair** — so once one non-`--as` bot interruption has happened, a *second distinct* bot's legitimate interruption is silently dropped (whereas at 3 it took more events). 
- The earlier "F1 persistent latch" note (counter only resets on a human proposer) now arms after 2 bot interruptions instead of 4. The previously-suggested "reset on queue drain" fix was analysed at threshold=3 and deserves a re-check at 1 before adopting — left as a separate item, not bundled here.

This remains an **automated preliminary** change; the maintainer's review is final and it is **not merged**.

Lotu527:
**Follow-up (automated, per maintainer review): fixed the session-global counter false-positive.**

Review feedback (2026-09-18): the earlier note flagged that the loop counter was a single session-wide scalar, so *"an unrelated bot's normal interruption also gets silently dropped"* — after one bot had interrupted, a **second, distinct** bot's first legitimate interruption was suppressed even though the two are not a ping-pong loop. At threshold 1 this fired after just one prior bot interruption.

Commit `b561bb1` on this branch keys the counter **per proposer identity** instead:
- Replaced the scalar `crossPrincipalConsecutiveBotInterruptions` with `crossPrincipalBotInterruptionCounts` (a `Record` keyed by `crossPrincipalProposerKey` = lark app + the bot's own union/open id; unattributable senders share one `unknown` bucket, kept consistent with `sameTrustedPrincipal`).
- `noteCrossPrincipalProposer` now takes the proposer's `TrustedCaller` and increments only that bot's tally; any non-bot (human/unknown) proposer clears the whole map, exactly as before.

Effect:
- **Fixed:** bot C's first interruption is no longer dropped because bot A looped — `count[C]` starts at 0, so C gets its one prompt.
- **2-body storm unchanged:** the owner only ever sees the single looping proposer, whose own tally climbs and trips.
- **N-body strictly better** than the global counter: each looping bot trips its own breaker independently, with no cross-bot collateral.
- **F1 (persistent latch) blast radius shrinks** from "all bots on the session" to just the offending bot.

Verified: `tsc --noEmit` clean; store suite 10/10 (added a regression test — bot A tripping does not suppress bot C's first interruption, mutation-checked: reverting the key to a global constant reds only that test — plus an unattributable-bot bucket case); daemon XPI suites (cross-principal-choice, ordinary-ingress-failure-notice, xpi-shared-cwd-admission / -daemon-callbacks, xpi-session-store-busy-retry) 54/54. The new field persists across restart via the same whole-object session JSON as before; the counts map is bounded by distinct bot identities (fleet-scale), not message volume, so no leak.

Still an **automated preliminary** change; the maintainer's review is final and it is **not merged**.

deepcoldy:
Thanks for the quick follow-up — both new commits look good. Automated preliminary re-review against the latest `origin/master` (rebased locally, no conflicts, `tsc`/build clean, XPI suites 64/64); maintainer review is final.

**Confirmed working (mutation-checked)**
- Threshold `3 → 1`: with per-proposer keying this means exactly one prompt per bot per run, silenced from its second interruption. The new literal test genuinely locks the value (changing the constant back to `3` reds it; the other cases still loop over the constant as expected).
- Per-proposer keying fixes a real false-positive the first version would have had at threshold 1: with a session-wide counter, a *second, unrelated* bot's first legitimate interruption would have been silenced merely because the first bot looped. Forcing the key back to one global bucket reds the new "distinct bot" test and only that test, so the isolation is really exercised. The two-body storm still terminates (each side only ever sees the one looping proposer), and N-body is strictly better.
- A non-bot proposer clearing the whole map is covered (removing the clear reds the reset test). The `--as` escape hatch is unaffected (settlement matches by proposer, ignoring suppression).
- We regard the earlier "persistent latch" concern as **resolved by this redesign**: the blast radius is now just the single looping bot, and at threshold 1 the only thing suppressed is the advisory repeat-prompt — the functional `--as` path always works and the first interruption always prompts.

**Please address before merge**
1. **Personal name in public history (repo policy).** A reviewer's real name appears in the body of two of the new commits and in a comment in `test/cross-principal-interruption-store.test.ts` ("Review feedback (…)"). This is a public repo and the project convention is no personal names in commits or source — please scrub them (a neutral "Review feedback, 2026-09-18" works). Note the in-source comment survives a squash merge, so that one definitely needs to go; we'll also write a clean squash message.

**Recommended (non-blocking)**
2. **Daemon wiring still has no behavioral test.** Forcing the staging suppress-gate to `false`, and forcing the drive-loop silent-removal to `false`, each leaves the full XPI cluster green (37/37). The pure counter is now well covered, but the two `daemon.ts` integration points aren't. The earlier four cases still apply via `__testOnly_driveCrossPrincipalInterruptions` (mocks in `daemon-ordinary-ingress-failure-notice.test.ts`): (a) 2nd same-bot interruption sends no notice and persists `loopSuppressed`; (b) drive loop removes it with no terminal; (c) a distinct bot's first interruption still prompts; (d) after suppression that bot's `--as` still settles.
3. **Wording nit.** The `daemon.ts` inline comment/log still say "enough consecutive bot proposers" / "resets the counter" — that's now per-proposer, so "consecutive interruptions from this proposer" / "clears every bot's tally" would match the new semantics (store/types docs are already updated).
4. **Tiny identity note.** `crossPrincipalProposerKey` prefers `unionId`, while `sameTrustedPrincipal` compares union only when both sides have one and otherwise falls back to open id. If one bot ever arrived with a union id on one message and without it on the next, its tally would split across two keys (throttling is only ever weakened by one prompt — never a false silence). Fine to leave; just be aware it isn't literally the same equivalence. The shared `unknown` bucket is defensive-only (the daemon always has at least one id at these call sites).

Net: mechanism is correct and ready; just the name scrub is a must, the daemon-wiring tests are the main worthwhile add.

PR #1426, opened 2026-09-16
fix(worker): gate standalone init on IPC readiness

xinna2001:
## Summary

- boot compiled standalone workers through a small entry module that installs the IPC preload before loading the CLI graph
- add a probe/ready handshake and hold standalone worker messages in FIFO order until the preload is listening
- start ordinary-IM delivery timers only after the real init is sent, with a bounded 30-second fallback
- leave the existing Node fork/preload path unchanged

## Root cause

The Node worker path injects `worker-ipc-preload.js` through `execArgv`, but the compiled standalone path re-execs the binary and necessarily drops `execArgv`. A large first-turn init payload can therefore hit IPC backpressure while the standalone worker is still loading. The daemon's two 2-second delivery attempts can expire before the full worker registers its message handler, so the first turn reports a delivery failure even though later turns work once the worker is warm.

## Verification

- `vitest run --project unit test/worker-ipc-preload.test.ts test/session-lifecycle-start.test.ts --maxWorkers=1`: 199 passed
- `npm run build`: passed
- compiled host standalone binary and ran `--version`: passed
- compiled-binary stress check with a 3-second stopped child and a 2 MiB init payload: init was released only after IPC readiness; send callback completed in 556 ms and worker receipt arrived in 558 ms

The full parallel unit run was not clean on this host because several long-running integration files timed out or hung under load. Two representative failures were rerun with `--maxWorkers=1` against both this branch and untouched `origin/master`; both reproduced identically on the base branch (the macOS `/var` vs `/private/var` worktree assertion and the TraeX native-title 14-second timeout).


deepcoldy (approved):
首审 + 复审均 0 阻断。已在最新主干（含 #1362）上本地 rebase 复验：无冲突、PR 净编辑逐字节不变、tsc build exit 0、重叠面 8 文件 644 通过（1 项 darwin-only 跳过）。编译态真机握手实测证实缺陷为真，作者自带测试变异对照 4/4 红。

PR #1453, opened 2026-09-18
fix(trigger): 精确终止被后继取代的交接

xluos:
新增需要认证的 `POST /api/sessions/:sessionId/trigger-result/supersede`：仅在明确指定的后继交接持久完成后，将同一会话中被取代的前驱标为终态，避免旧交接持续 pending。调用方必须自行证明因果关系，不按时间先后推断。

后继需要完整的完成记录；前驱须为 pending 或不确定的 dispatch_unknown。校验机器人归属，保留其他交接，重复请求幂等，并及时清理内存中的旧 pending。后来的真实 worker 完成或终止证据仍可覆盖标记。接口不加入 core-only 免认证列表。

带 `steerParkedBy` 的 pending 前驱返回 `409 / predecessor_steer_parked`。判断位于幂等短路之后，保留原记录和指针，由 steer 链负责其最终结果，保证重启后仍能读取合并完成结果。已同步 master 的 steer 实现并解决相邻新增函数的冲突。

工具子进程收到非敏感的会话、群和话题路由标识。修复环境变量测试的源码定位锚点及断言，保持会话 ID 参数唯一且顺序稳定，两种相关 CLI 适配器均跳过 undefined 值。

影响面：共用 IPC 与持久化路径覆盖群、话题及无界面会话；认证、归属检查和持久完成优先级保持有效。路由变量不包含凭证。

验证（Linux 隔离源码目录，Node 24.21.0 / Bun 1.4.2）：

- `bun run test test/async-trigger-store.test.ts test/trigger-user-shell-env.test.ts test/codex-read-isolation.test.ts test/cli-adapters.test.ts`：513 项通过。
- `bun run test test/dashboard-ipc.test.ts`：276 项通过、1 项跳过；覆盖鉴权、409、200、幂等以及后续 GET 返回 failed。
- `bun test test/async-trigger-store.test.ts test/trigger-user-shell-env.test.ts`：51 项通过。
- 新增存储用例验证 steer 前驱拒绝且文件内容不变、指针保留、链解析 completed，以及普通 pending 的替代与幂等。
- `bun run build`：通过。

测试中的外部服务由替身隔离，尚未执行真实群聊验收。

CI 稳定性修复（`1be517f7`）：预览代理测试原先先释放临时端口再启动前端，存在端口被复用的竞态，可能将预期 502 变成 404。已改为持续占用端口、收到连接立即关闭的 TCP 测试服务，并在 finally 中清理。强制端口复用可复现旧失败；修复后 Bun/Node 的预览代理测试各 36 项通过，目标用例 4 进程并行重复 100 次全部通过。


deepcoldy:
你好，这是 botmux 的自动评审流程：我们已为这个 PR 建立了飞书评审群，首次 review 与复审都会在群内进行，最终结论也会同步在这里。

请点击下面的链接加入评审群（一年有效）：
https://applink.larkoffice.com/client/chat/chatter/add_by_link?link_token=65dq51c5-8e05-497c-87ae-be3febe72fad

另外，你目前还不在我们的「GitHub 账号 ↔ 飞书」自动拉群名单里，可以顺手在这个名单文档中补上你的 GitHub 账号和飞书信息，之后的 PR 评审群就会自动把你拉进去：
https://bytedance.larkoffice.com/wiki/WJ1nwWbtxi89erkNGNbcgkt9nUe

（本条为自动流程消息，敬请忽略措辞上的不周 🙏）

deepcoldy:
Thanks for the PR — the strict-exact-supersession design and the routing-env injection both look solid (owner fail-closed checks, strict loader + file lock + fsync, HMAC-required route kept out of the core-only public allowlist, idempotency, no secrets in the shell env). We rebased the branch locally onto the latest `master` (`be1b42403`), clean rebase, and the new unit tests pass with mutation checks.

One blocking issue — it is a test-only regression introduced by this change, which is why the CI `test (2/3)` and `bun-test` legs are red.

## 🔴 Blocking: source-sliced test anchor broken (CI shard 2/3 red)

`test/trigger-user-shell-env.test.ts` slices `src/worker.ts` between two hard-coded source anchors at module-load time:

```ts
const start = worker.indexOf('const identityShellEnv: Record<string, string> = {};');
```

This PR changes that line into a multi-line object literal, so `indexOf` returns `-1` and the whole suite throws at load (`Cannot locate worker identity environment assembly`, 0 tests). The file hashes to vitest shard 2/3, which matches the red CI leg; `bun test test/trigger-user-shell-env.test.ts` fails the same way locally.

The fix is not just the anchor — once it locates the block again, the three exact `toEqual` assertions in that file go red, because routing env vars are now injected unconditionally:

1. Anchor → `'const identityShellEnv: Record<string, string> = {'` (without the closing brace).
2. The `workerShellEnv` cfg fixture currently passes only `{ sessionId, triggerUserAuth }`; add `chatId` + `larkAppId` (and `rootMessageId: 'om_…'` if you want a thread-scope case).
3. Update the exact-equality assertions:
   - the enabled-auth case (`~line 51`) gains `BOTMUX_CHAT_ID`, `BOTMUX_LARK_APP_ID`, `BOTMUX_SESSION_SCOPE: 'chat'` alongside `BOTMUX_SESSION_ID`;
   - both `toEqual({})` cases (`~lines 62/63`, auth-off / no-data-root) now also expect the four routing keys — the env is assembled regardless of `triggerUserAuth`, only the identity-wrapper paths remain conditional.

Verified locally that with those adjustments the file goes green.

## Non-blocking suggestions

- **Duplicate codex `-c` arg**: `codex.ts:256` already pushes `shell_environment_policy.set.BOTMUX_SESSION_ID=…`, and the `shellSubprocessEnv` loop now adds it a second time (`buildArgs` produces two identical `-c` entries — verified). Harmless (last wins) but redundant; consider removing the key on one side.
- **`undefined` defense in the shell-env loop**: `JSON.stringify(undefined)` interpolates to a bare, unquoted `…=undefined` (invalid config value). Unreachable today (`cfg.chatId` is required), but an `if (value === undefined) continue` guard in the adapter loop would make the unconditional injection future-proof. Same applies to the traex adapter.
- **Successor shape check**: the supersede transaction only checks `successor?.status === 'completed'` without the strict `isValidPersistedResult` shape validation — a hand-corrupted file with `{status:'completed'}` but no `completedAt` is accepted as proof. No in-repo writer produces that, so defense-in-depth only.
- **Live in-memory state**: superseding writes disk but leaves the live `ds.asyncTriggerResults` entry as `pending`. Polls already resolve correctly (the disk `turn_terminal` ranks above in-memory pending), and it self-heals on worker terminal; deleting the in-memory entry in the same path would converge immediately.
- **Behavioral test coverage for the new IPC route**: currently only the 401 allowlist guard is asserted. Happy path (`200` / `alreadyTerminal`), `409` (`successor_not_completed` / `predecessor_not_pending`), and a follow-up GET converging to `failed` would be worth adding.
- A late real `completed` (or a late worker terminal failure) deliberately overwrites the supersede marker — consistent with the store's existing completed-wins contract; a comment on the function saying the marker is not resilient to later strong evidence would make that explicit.
- Optional: document the new endpoint and the `BOTMUX_*` routing shell vars in `docs/dispatch-receipt.md`.

🤖 This is a preliminary automated review (two independent passes); the final call is of course up to the maintainers. Happy to discuss any of the above.

deepcoldy:
Follow-up review after the two new commits (`54dba48`, `0465869`) — rebased locally onto the latest `master` (now `f50012a`, which added #1445 steer support).

## Previously blocking item: fixed ✅

The sliced-test anchor is updated together with the cfg fixture and all three exact-equality assertions, and the renamed test reflects the new "routing env always present" contract. The dedup/undefined assertions cover the adapter side. CI is now 9/9 green (including shard 2/3 and the advisory bun-test leg), and the full shard 2/3 suite passes locally apart from one pre-existing, root-only environment failure (`mojo-launcher-env-quarantine`, chmod-based EACCES assertions that don't trigger under root; green on CI's non-root runner).

All six non-blocking suggestions from the first review are addressed:
- the codex duplicate `-c` is gone (verified: exactly one `BOTMUX_SESSION_ID` override, position/order preserved, and it now carries the worker's env value — which also aligns it with `cfg.sessionId` on resume);
- `undefined` values are skipped in both codex and traex loops (verified no bare `=undefined` is emitted);
- the successor now goes through the strict shape validator, with a corruption test;
- the live in-memory pending entry is deleted on success;
- the new IPC behavioral test covers 409 → 200 → idempotent 200 → follow-up GET `failed` → missing-predecessor 409, over a real HMAC-signed request;
- marker semantics are documented in the function comment and `docs/dispatch-receipt.md`.

## One new interaction surfaced by the rebase onto #1445 (non-blocking, worth a decision)

#1445 added a second "predecessor parked behind its successor" mechanism: a codex-app steer-group member stays `pending` with a durable `steerParkedBy` pointer, and after a restart the GET handler walks that chain and mirrors the group's **merged final answer back as `completed`**. The two mechanisms don't currently coordinate:

- Calling this endpoint on a `pending` predecessor that carries `steerParkedBy` rewrites it to `failed/superseded_by_completed_successor:<succ>` (the pointer is not preserved), so the chain-walk precondition (`status === 'pending' && steerParkedBy`) no longer fires — after a daemon restart that member polls `failed` instead of the merged success #1445 promises.
- Without a restart, `fanOutSteerGroupFinal` unconditionally calls `recordCompleted` for every parked member, so the marker is overwritten back to `completed` (completed-wins). The terminal state then differs between live and post-restart depending on timing.
- A caller cannot tell the two cases apart: `trigger-result` never exposes `steerParkedBy`.

Reproduced locally with parked `T1→T2` + `T2` completed: supersede returns `superseded`, `T1` ends `failed` with no park pointer. Narrow reach (an authenticated caller superseding a steer-group member specifically, plus the restart window), no security or data-loss impact, but the two contracts disagree on the same row. Cheapest option: in the store transaction, reject a `pending` predecessor that carries `steerParkedBy` with a distinct outcome (e.g. `predecessor_steer_parked`), leaving its terminal state exclusively to the steer machinery — plus one test. Alternatively, explicitly document that superseding a steer member is intended and the failed state wins after restart.

Minor aside: after the live in-memory pending entry is deleted, a still-running predecessor that emits a late `final_output` no longer takes the async-result branch. The codex-app path already fail-closes this (`codexAppDeliveryMustFailClosed`); the generic HTTP-async path would attempt the ordinary Lark reply toward the synthetic `http_…` chat id, which fails harmlessly (best-effort log, no external send). Suppressing that case symmetrically would be a tiny cleanup.

🤖 Preliminary automated review (two passes + rebase cross-check); final call remains with the maintainers.

deepcoldy:
Update from maintainer review: the design is accepted; before merge we'd like the one interaction item from the previous comment fixed — the `steerParkedBy` coordination with #1445. Concretely:

1. In `supersedePendingTriggerByCompletedSuccessorStrict`, **after** the `already_superseded` idempotency short-circuit and **before** the `predecessor_not_pending` check, reject a `pending` predecessor that carries a `steerParkedBy` pointer:

   ```ts
   if (predecessor?.status === 'pending' && predecessor.steerParkedBy) {
     return 'predecessor_steer_parked';
   }
   ```

   No disk write — the row stays `pending` with its pointer so the #1445 steer chain-walk remains the sole owner of that turn's terminal state. (Placing it after the idempotency branch matters: an existing supersede marker must still return `already_superseded`.)

2. Add `'predecessor_steer_parked'` to the `SupersedePendingTriggerOutcome` union. A distinct code is preferred over reusing `predecessor_not_pending` — the predecessor genuinely *is* pending; callers should not misread this as "already terminal". The IPC route already maps every non-success outcome to `409`, so no handler change is needed, but the response will now carry the precise error code.

3. One store-level test: a parked predecessor (`recordSteerParked(M→S)` + `S` completed) is rejected, remains `pending` with its pointer, and `followSteerParkedChain(M)` still resolves to the merged `completed`; plus an assertion that an ordinary (non-parked) pending predecessor still supersedes normally.

Please also rebase onto the latest `master` when you push (the branch currently conflicts with #1445 in `async-trigger-store.ts` — purely adjacent additions, trivial to resolve by keeping both).

Once CI is green we'll merge. Thanks again for the quick iteration on the first round.

🤖 Automated review coordination; the direction above reflects the maintainer's decision, final merge is handled by the maintainers.

xluos:
@deepcoldy 已按[这条评审意见](https://github.com/deepcoldy/botmux/pull/1453#issuecomment-5743321637)修改并推送，最新提交为 `203a2adf`：

- 在 `already_superseded` 幂等短路之后、`predecessor_not_pending` 判断之前，拒绝覆盖带 `steerParkedBy` 的 pending 前驱，返回新增的 `predecessor_steer_parked`；IPC 沿用 409 映射。
- 拒绝时不写盘，保留 pending 和指针。新增测试验证文件内容不变、`followSteerParkedChain` 仍解析为合并后的 completed，以及普通 pending 仍可正常 supersede、重复请求仍幂等。
- 已 rebase 到最新 master 并解决冲突，补充接口文档。

Linux 定向验证：Vitest 789 项通过、1 项跳过；Bun 51 项通过；`bun run build` 通过。

另外同步当前 CI 状态：8/9 检查通过，`bun-test` 有一项失败，位于 `test/session-preview-proxy.test.ts` 的不可达 HTTP/WebSocket 目标用例（预期 502，实际 404）。原因尚未确认，当前还不能报全绿。

感谢指出 steer 完成链的交互问题，麻烦再看一下这次修改。

deepcoldy (approved):

PR #1490, opened 2026-09-21
refactor(e2e): migrate Feishu scenarios to Midscene Test

quanru:
## Summary

- Migrate all 17 existing Feishu browser scenarios from the Vitest lifecycle to native Midscene Test YAML registration.
- Preserve the original Playwright and Midscene Web scenario bodies, setup, teardown, session cleanup, and schedule cleanup behavior.
- Use aiAct for visible business interactions; keep aiWaitFor and aiAssert for observable state and verification, following the flow-web-monorepo convention.
- Add a credential-free Dashboard smoke project that always executes a real Midscene case and produces a native report, even when Feishu credentials are unavailable.
- Route test:e2e-browser through the Midscene Test runner and keep run data below midscene_run.
- Add a dedicated Midscene E2E workflow that uploads the complete report artifact, writes a real case Summary, and publishes the native HTML report to GitHub Pages.
- Add generated node documentation, focused type checking, and registry validation.
- Exclude the migrated browser scenarios from Vitest so test:all does not execute them twice.
- Upgrade Midscene Web and add Midscene Test at version 1.13.0.
- Keep the lockfile on the public npm registry for GitHub Actions.

## Why

Botmux already had substantial Midscene-based Feishu browser coverage, but it was orchestrated by Vitest. This change moves those existing cases under Midscene Test while preserving their behavior. The separate Dashboard smoke project provides a credential-free CI path so every same-repository pull request produces an actual Midscene execution, Summary, artifact, and HTML report instead of reporting a fully skipped run as success.

## Impact

This is test infrastructure only; production runtime behavior is unchanged. The only unrelated test adjustment widens the process-level statusline watchdog assertion from 12 to 15 seconds while keeping the production watchdog at 10 seconds, accounting for process startup and shutdown observed in GitHub Actions.

The Feishu project still requires its test-group URL and authenticated Playwright storage state. When they are unavailable, only those 17 live cases are skipped. The Dashboard project and report generation remain required. The four Midscene model secrets are configured in the contributor fork.

## Validation

- Real Midscene run, report artifact, Summary, and Pages deployment passed: https://github.com/quanru/botmux/actions/runs/35594161795
- Published native Midscene HTML report: https://quanru.github.io/botmux/
- Pull-request Midscene workflow passed: https://github.com/quanru/botmux/actions/runs/35594153089
- Dashboard Midscene project: 1/1 case passed, 7/7 steps passed.
- Feishu scenario registry: 17/17 mappings validated.
- Both Midscene project type checks passed.
- Both Midscene node registries generated successfully.
- Midscene registry unit test passed.
- Pages report packaging and HTTP endpoint were verified.
- git diff --check passed.

The complete repository CI passed on the fork after rerunning one unrelated flaky systemd/tmux scope test: https://github.com/quanru/botmux/actions/runs/35594152936
## Skipped-case CI example

The fork intentionally omits the Feishu group URL and authenticated browser state in this example. The workflow still runs the Dashboard Midscene case and keeps every unavailable Feishu case visible instead of dropping them from the report:

- Example CI: https://github.com/quanru/botmux/actions/runs/35596623940
- Published report: https://quanru.github.io/botmux/
- Result: 1/18 passed, 0 failed, 17 explicitly skipped
- Complete report artifacts: botmux-midscene-report and botmux-midscene-pages-1


deepcoldy:
你好，感谢 PR！我们的自动评审群已建好，欢迎点击下面的链接加入飞书群，评审意见和后续沟通都会在群里同步：

https://applink.larkoffice.com/client/chat/chatter/add_by_link?link_token=261jbf58-a8fb-4d0a-a0ca-ee8d794c13d9 （一年有效）

目前自动拉群名单里还没有你的飞书信息，所以暂时无法直接把你拉进群。也可以把 GitHub 账号和飞书信息补进这个名单文档，补好后后续评审会自动拉你入群：https://bytedance.larkoffice.com/wiki/WJ1nwWbtxi89erkNGNbcgkt9nUe

本条消息为自动流程发出，最终评审结论以维护者审阅为准。

deepcoldy:
Thanks for the clean migration! We reviewed it end to end — 17/17 registry resolution, YAML ↔ zod-enum ↔ scenario-target cross-checks, the `@midscene/test@1.13.0` runner/CLI/summary contracts, and a lockfile audit. **No blocking issues**: the scenario bodies, setup/teardown hooks, and schedule cleanup behavior are preserved, and everything stays in `devDependencies` (no production/runtime impact).

A few **non-blocking** notes (for the PR description and/or a follow-up). This is an automated first-pass review; the maintainer's review is final.

**A. The orphan-schedule sweep moved — worth stating as a behavior change.** `scripts/run-e2e.ts` no longer sets `BOTMUX_E2E=1`. The stale-task cleanup that previously ran from the Vitest `globalSetup` now runs by calling `sweepOrphanSchedTasks()` directly inside `run-e2e` before spawning the runner. Nothing else sets `BOTMUX_E2E` anymore, so `test/global-setup.ts` is effectively a backward-compatible no-op. It works correctly; one line in the PR description would stop future readers from assuming the Vitest global setup still performs the sweep.

**B. Dependency blast radius (devDependencies only).** `@midscene/web` moves 1.12.1 → 1.13.0 and `@midscene/test@1.13.0` is added. The new package pulls `glob@11`, `yargs@17.7.2`, `get-stream@9`, `signal-exit@4`, `tinyglobby@0.2.15`, which flip the top-level hoisted entries (`glob` 7→11, `yargs` 17.7.3→17.7.2). The desktop-build chain keeps its versions as nested copies (`@electron/asar/glob@7.2.3`, `rimraf/glob@7.2.3`, `electron-builder/yargs@17.7.3`, `@puppeteer/browsers/yargs@17.7.3`); `electron-builder@26.15.3` declares `yargs: ^17.6.2` (both 17.7.2/17.7.3 satisfy it) and does not depend on `glob` directly. Nothing here ships in the artifact, and main CI does not run electron packaging — but a sentence in the description would be useful.

**N4. Consider pinning `tsx` to avoid a latent runner break (fine as a follow-up).** The runner loads the project through midscene-test's internal `require('tsx/esm/api')`; that import breaks on tsx 4.23.15+ (it resolves to a non-existent `dist/esm/api/esm/`), while 4.23.12–4.23.14 work. The lockfile currently resolves a single `tsx@4.23.12` — the repo's own `tsx: ^4.20.5` wins and satisfies @midscene/test's `^4.19.2` — so CI is safe today. A future dependency refresh could silently move to 4.23.15+. Pinning the repo's own devDependency `tsx` to `~4.23.14` (or an exact `4.23.12`) would also constrain the runner without waiting on upstream.

Optional nits:
- `run-e2e.ts` spawns `midscene-test` via PATH (`shell: false`). It's fine under `bun run`, but resolving the bin explicitly (or re-execing via `process.execPath`) would make a bare `tsx`/direct invocation robust.
- A handful of user-facing hints still read `pnpm test:e2e-browser:setup` (in `setup-login.ts`, `bot-test-factory.ts`, several e2e error messages, and the report-dashboard empty state). Switching them to `bun run …` would match the migrated tooling.

None of this blocks merging — A/B can also be folded into the squash notes, and N4 is a reasonable post-merge follow-up. Thanks again for the migration!

deepcoldy:
Thanks for following this up — the Dashboard smoke project and Pages publishing are a nice addition. We re-reviewed the expanded diff at `477b35e31` (rebased onto current `master`). Because the PR grew well beyond the original Vitest→Midscene shell migration, our earlier non-blocking assessment no longer covers it. **We'd suggest addressing the following three before merge.** This is an automated first-pass review; the maintainer's review is final.

**1. The 60-minute job timeout is shorter than the Feishu budget, and a timeout loses the report.**
`midscene-report` is a single job with `timeout-minutes: 60` (previously 180) that covers install + Chromium + build + Dashboard + the 17 Feishu cases. The Feishu cases run with `maxConcurrency: 1`, `retry: 1` on CI, and no early bail (`bail` defaults to 0); the per-case timeouts declared in the `.e2e.ts` files sum to ~122 min (and note those per-case timeout arguments are inert under the new mini registry — the actual per-case ceiling is the Feishu project's 15 min `testTimeout`). On top of that, the summary/upload/Pages steps are all gated `if: always() && !cancelled()`, so when GitHub kills the job at 60 min it becomes `cancelled` and those steps are skipped — no summary, artifact, or Pages report is produced.
Suggested fix: split the quick, model-config-gated Dashboard run from the slow, secret-bearing Feishu run into **two jobs with independent timeouts** (this also keeps Dashboard feedback from being held hostage by the long Feishu run). Restoring a ≥180-minute budget would also work.

**2. "always runs the credential-free Dashboard project" isn't accurate — please align the wording (or make it truly credential-free).**
This phrasing appears in the workflow/README and in `render-midscene-summary.mjs`, but (a) the "Require/Check Midscene model configuration" steps are unconditional, and (b) `dashboard-smoke.yaml` uses seven `aiAssert`/`aiAct`/`aiWaitFor` steps that require a vision model. The trigger is `pull_request` (not `pull_request_target`), and `MIDSCENE_MODEL_*` are repository secrets that GitHub never injects for fork PRs — so on this fork PR both Midscene jobs show `skipping` and the Dashboard project has not actually run in CI. Splitting jobs (item 1) lets Dashboard run independently on same-repo branches and `workflow_dispatch`, but it still can't run on fork PRs. Either soften the wording to something like "runs when model configuration is available; skipped on fork PRs", or replace the `ai*` steps with deterministic assertions if a genuinely credential-free gate is the goal.

**3. The Pages/Summary scripts structurally can't see Feishu results and report them as skipped even when they pass.**
`render-midscene-summary.mjs` only receives a string `--feishu-outcome` (and `--results` points at the Dashboard result dir), and `prepare-midscene-pages.mjs` only gets the Dashboard `--results-root` plus `--skipped-cases-dir`. Neither receives the Feishu run's `summary.json`, so `prepare-midscene-pages.mjs` unconditionally lists all 17 Feishu cases as skipped (and even throws when there are zero skipped cases). In a real secret-bearing run where all 17 Feishu cases pass, the published Pages landing page still reads `1/18 cases passed · 17 skipped`. The current test (`test/prepare-midscene-pages.test.ts`) asserts this behavior (`1/3 … 2 skipped`) and has no "Feishu actually ran/passed" branch.
Suggested fix: pass the Feishu result directory into both scripts, only synthesize skipped rows when Feishu is genuinely skipped, and add a test for the Feishu-ran/passed path (the existing assertion will need updating).

Non-blocking:
- `findReportIndexes` only matches a file literally named `index.html`; it happens to work with Midscene 1.13's directory-mode report, but accepting any `*.html` would be more robust.
- The change is no longer strictly "test bodies unchanged": `helpers.scrollThreadToBottom` and `feishu-group-mention.e2e.ts` replace three `aiScroll(undefined, { scrollType: 'untilBottom' })` calls with `aiAct('…scroll to bottom…')` (presumably a 1.12→1.13 compatibility fix). That's fine, just worth calling out in the description. The committed `test/midscene/midscene-node-reference.md` is a ~2,500-line generated file pinned to 1.13 — fine now, but it will drift on the next upgrade.
- The earlier note still applies: pinning the repo's own `tsx` to `~4.23.14` would keep `@midscene/test`'s internal `tsx/esm/api` usage off the 4.23.15+ break; a reasonable follow-up.

None of these touch production code — they're all about making the new CI/reporting pipeline honest and reliable in the secret-bearing runs it exists to serve. Thanks again!

quanru:
Superseded by #1512, which runs the complete Midscene suite from an upstream repository branch with access to the required Actions secrets.

